PR #31, opened 2018-04-26
convention: GROUPID == pcGtsId

kba:
For discussion, not a hard requirement but I think it would make sense to recommend this.

Came across this while pondering how to add extracted line images to METS fileGrp.


cneud:
I think this is indeed a very "nice to have" for a recommendation.

Issue #30, opened 2018-04-26
ocrd-tool: Coupled with cli?


kba:
If `ocrd-tool.json` is independent from the CLI (as it is now) the only link between the tool and its description is via the binary name. This has drawbacks:

  * Discovery: If a tool `ocrd-xyz` is installed there is no standardized way to get e.g. the parameter description.
  * Consistency: The description about a tool and the actual tool might diverge, e.g. changed parameters, change name etc.
  * Duplication: When the tool description isn't actually part of the tool, the tool itself cannot access its own description, e.g. default values for parameters.

Maybe make that part of the CLI?

```json
ocrd-xyz --tool-desc
{
  "repo_url": "...",
  "parameter": {
    "threshold": {"type": "integer", "default": 0.2}
  }
}
```

VolkerHartmann:
You want the tool to return the complete content of ocrd-tool.json?
I think that's a really good idea, but it's a kind of chicken egg problem, isn't it.
We need at least the executable in beforehand.
We already defined a standard path and a standard filename for the file.
'Tools MUST be described in a file ocrd_tool.json in the root of the repository.'
In my opinion we 'only' have to register the URL of the repository and if there are any changes
we must have the possibility to rescan the repo and/or to version the algorithm by its hash (I think tag may not enough)

kba:
Add to CLI:

> * `--ocrd-tool, -J`: Return `ocrd-tool` description for this command, in particular its parameters

How MP organize this is up to them, but I found easiest:

* Have `ocrd-tool.json` somewhere in the distribution of the software and accessible from code
* Make a symlink to that file in the repository root

kba:
Added in 1.2.0

PR #47, opened 2018-05-22
add -J/--ocrd-tool option as per #30

kba:
see https://github.com/OCR-D/spec/issues/30#issuecomment-389862850


kba:
Can be merged.

PR #56, opened 2018-06-13
Glossary

kba:


cneud:
I found some smaller things that require fixing, working on it.

kba:
@cneud Thanks, updated on the site.

PR #63, opened 2018-06-18
require ocrd-tool.json to have a version, OCR-D/core#109, OCR-D/spec#…

kba:
…51, OCR-D/ocrd_kraken#13


kba:
No, order doesn't matter.

PR #66, opened 2018-07-20
convention how URL are to be interpreted in the CLI

kba:
In the CLI, URLs should be handled more flexibly.

In addition to

```sh
ocrd-compliant-tool -m file:///path/to/mets.xml
```

allow these equivalent shortcuts:

```sh
ocrd-compliant-tool -m /path/to/mets.xml
cd /path/to
ocrd-compliant-tool -m mets.xml
```

Same whereever a URL is expected.

The [OCR-D core framework](https://github.com/OCR-D/core) supports this behavior.

### DIFF
--- a/cli.md
+++ b/cli.md
@@ -36,7 +36,7 @@ The `mets:file@GROUPID` to process. If no groupID is defined all files within th
 
 ### `-p, --parameter PARAM_JSON`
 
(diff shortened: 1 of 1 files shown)


VolkerHartmann:
:+1:

PR #74, opened 2018-09-14
Sharpen specification due to multiple values and logging.

### LINKED ISSUE #68: CLI: Specifying multiple input/output groups
Specs [say](https://ocr-d.github.io/cli#-i---input-file-grp-grp)

> `-I, --input-file-grp GRP`
> File group(s) used as input.

but not how multiple groups are to be given, right?

The most straightforward convention would be to use comma, as for `--group-id` and as implemented in [cis-ocrd-align](https://github.com/cisocrgroup/cis-ocrd-py).

VolkerHartmann:
Fix #68

### DIFF
--- a/cli.md
+++ b/cli.md
@@ -1,6 +1,8 @@
 # Command Line Interface (CLI)
 
-**NOTE:** Command line options cannot be repeated. To specify multiple values, provide a single string with comma-separated items (e.g. -I group1,group2,group3 instead of -I group1 -I group2 -I group3).
+**NOTE:** Command line options cannot be repeated. To specify multiple values,
+provide a single string with comma-separated items (e.g. `-I
+group1,group2,group3` instead of `-I group1 -I group2 -I group3`).
 
 ## CLI executable name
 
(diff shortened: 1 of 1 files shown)


cli.md
```diff
@@ -1,5 +1,7 @@
 # Command Line Interface (CLI)
 
+:information_source: If a parameter supports multiple values they have to be specified via a comma separated list.
```
kba: Command line options cannot be repeated. To specify multiple values, provide a single string with comma-separated items (e.g. `-I group1,group2,group3` instead of `-I group1 -I group2 -I group3`).

kba: Better not use emoji short names, they won't be rendered in the final output. I use `**NOTE:**` for sidebars which isn't as pretty but clear.

cli.md
```diff
@@ -41,6 +43,7 @@ URL of parameter file in JSON format
 ### `-l, --log-level LOGLEVEL`
 
 Minimum Log level. One of `OFF`, `ERROR`, `WARN`, `INFO` (default), `DEBUG`, `TRACE`).
+:warning: This specification overwrites all other specifications. For a fine-grained change, please use the parameter file.
```
kba: What is the `parameter file`? And what if both are provided? Does this mean that when the command line flag is set, that it will override all other mechanisms? Even if those other mechanisms specify a lower or higher threshold for a certain area of the application? See https://github.com/OCR-D/core/issues/173

cli.md
```diff
@@ -42,8 +42,8 @@ URL of parameter file in JSON format
 
 ### `-l, --log-level LOGLEVEL`
 
-Minimum Log level. One of `OFF`, `ERROR`, `WARN`, `INFO` (default), `DEBUG`, `TRACE`).
-:warning: This specification overwrites all other specifications. For a fine-grained change, please use the parameter file.
+Set the global minimum Log level. One of `OFF`, `ERROR`, `WARN`, `INFO` (default), `DEBUG`, `TRACE`).
```
bertsky: Sorry, this may come too late, but I think it should really read `maximum` instead of minimum here. (Since each level has to let lower levels through, respectively, and the usual semantics of this is "how much information to display".)

bertsky: Yes, that would be clear, too. (Though I believe "log level" usually has more of a "verbosity" than a "severity" connotation.)

VolkerHartmann: I found this in the documentation of Log4J:
https://logging.apache.org/log4j/2.x/manual/customloglevels.html
Off (0), FATAL(100), ERROR (200),...
--> Set the global maximum verbosity level. More verbose log entries will be ignored.

kba: Fixed in a88c3ac

Issue #77, opened 2018-09-20
OCR on line vs word level


bertsky:
The current [specification](https://ocr-d.github.io/glossary#OCR) is agnostic about which **level of segmentation** OCR is supposed to operate on, either `TextLine` layout input (for `TextLine`, `Word` or `Glyph` text output), or `Word` layout input (for `Word` or `Glyph` text output). Arguably, this is a question of _workflow_ configuration. 

But if formatting / **font features** are annotated (as essential part of the workflow) up to the `Word` level – see #76 –, then OCR will need to proceed on the _word level_, too: To make use of the (word-level) font features in its input (improving recognition), and to be able to match the previous (word-level) annotation in its output (so font features can be re-used for post-correction or by applications). (Simply creating a new annotation with different `Word` elements would make it hard if not impossible to re-combine the different PAGE annotations.) 

On the other hand, if OCR does not proceed on the _line level_, its results will probably be suboptimal: Engines want to be free to choose the best word segmentation based on their own models, including dictionaries and language models. In order to still apply OCR-internal **language modelling** at all with word-level recognition, their API would need some way of pointing them to the context of previous results, or become stateful.

Furthermore, when multiple OCR results are **aligned** for post-correction, _allowing_ word segmentation ambiguity can actually help – see #72. But that requires staying on the line level. If instead OCR lines are aligned on the word level (i.e. preserving any one of its `Word` segmentations, deemed "master OCR" by CIS), then wrong segmentation will stay uncorrected. But then at least word-based font features can be used as input and kept in the output.

We already had a related [discussion](https://github.com/OCR-D/ocrd_tesserocr/pull/17) on the correct behaviour of the Tesseract wrapper (internal vs external `Word` layout), but I am afraid this issue is bigger. Or am I mistaken, are there other options? Can we at least get a clearer picture of this?

kba:
Algorithms should declare their "level of operation", whether they OCR on lines or on words, or merge OCR results under the assumption that all input has the same word segmentation or not.

@tboenig and I had some discussions about "Erfassungstiefe" to annotate PAGE and METS with. In essence, how to succinctly express that "this PAGE file has been segmented until word level with font annotations on the word level but without actual OCR" or "this METS file comes with logical page-spanning structure information".

[...]

Issue #72, opened 2018-08-14
allow intermediate PAGE annotation for word segmentation ambiguity


bertsky:
In the OCR-D workflow, there are several steps that likely require input or output to be able to represent __word segmentation ambiguity__ and confidence values of word boundaries (whitespace characters):

1. Each individual OCR engine might be able to provide such information. Postcorrection could benefit from it (importing whitespace characters with confidence as but one of many alternative characters within the input graph). Tesseract LSTM models definitely have this, at least internally (see OCR-D/ocrd_tesserocr#7 which is distinct but related). PAGE output needs to incorporate those without loss.

2. Character-based language models need whitespace characters with confidence attribute both for input and for output. 

3. Alignment of multiple OCR results produces symbol pairings including the empty symbol (insertion/deletion) and whitespace symbol (tokenization ambiguity) for each line. Since there is no reserved character for the empty symbol within `TextLine:TextEquiv:Unicode` – all of Unicode is possible here, except control characters, which are forbidden in CDATA –, we cannot use it to encode such alignments. We do have a natural representation for empty symbol at `Glyph:TextEquiv:Unicode`, but Glyph already necessitates a strict (hierarchical) Word segmentation, which would break tokenization ambiguity again.

Thus, it seems advisable to allow a PAGE annotation as interface at those particular steps which deviates from the standard in that 
* it avoids the strict word segmentation of `WordType` (_by using only 1 Word per Line per convention_), 
* but instead allows flexible word segmentation via `GlyphType` (_by using whitespace within its TextEquiv_). The alternative would be to either not use PAGE at all there or loose important information by design.

Example for multi-OCR alignment:
```XML
<TextLine>
 <Word> <!-- only 1 Word per TextLine -->
  <Glyph> <!-- glyph segmentation ambiguity within OCR (alignment by non-uniform length) -->
   <TextEquiv index="0" conf="0.9" dataType="ocr1">m</TextEquiv>
   <TextEquiv index="1" conf="0.8" dataType="ocr1">ni</TextEquiv>
   <TextEquiv index="2" conf="0.4" dataType="ocr1">ri</TextEquiv>
   <TextEquiv index="3" conf="0.3" dataType="ocr1">rn</TextEquiv>
   <TextEquiv index="4" conf="0.7" dataType="ocr2">n</TextEquiv>
   <TextEquiv index="5" conf="0.6" dataType="ocr2">r</TextEquiv>
  </Glyph>
[...]

PR #82, opened 2018-10-09
PAGE: How to add TextEquiv and consistency rules, OCR-D/assets#16

kba:
Should be mostly obvious, but feedback appreciated on wording and handling whitespace (current proposal is to use nbsp since most string stripping algos do not include it by default and yet it is Unicode `WHITESP=Y`).

@bertsky @finkf


bertsky (changes requested):
Thanks for going forward with this! IMO the overall direction of the language is right, but there are a few points with room for improvement:

  - [X] When we attach text recognition results, we might have multiple variants/hypotheses. Annotating those should be encouraged. Each one should go into a separate `T` with ascending `@index` and (then mandatory) descending `@conf`. 
  - [ ] Since this is the intended semantics of the `TextEquiv` sequence, we should forbid other uses within one PAGE. So either there is only one `T` or a list of alternative hypotheses with differentiating attributes.
  - [X] I find NBSP is a good idea for this corner case. The second sentence could be made more clear though by inserting "at the start or end of some `T`" before the comma.
  - [X] Terminologically, I think we should stick to "recognized text" or "text results" throughout and ditch "text equivalence".
  - [X] I would try to avoid the possessive form of XML identifiers. How about: "The text of each `<pg:Word>` must be equal to the texts of all `<pg:Glyph>` contained by it, concatenated directly."
  - [X] Maybe (just to be as clear as possible) one should even mention that concatenation adds strings in between but never leading (first) or trailing (last).
  - [ ] There might be valid or even necessary exceptions to this general principle in corner cases. Before we make this a hard requirement, we should search for those cases in our data semi-interactively. See [this comment](https://github.com/OCR-D/assets/issues/13#issuecomment-422069829)
  - [X] As far as I understood @tboenig in [this comment](https://github.com/OCR-D/assets/issues/12#issuecomment-422020664), failing the consistency assertion would _not_ allow a processor to stop, but require it to proceed with the content of the top-most compliant hierarchy level.

kba:
@bertsky All valid and helpful, thank you. I will adress them ASAP and update the PR.

> As far as I understood @tboenig in this comment, failing the consistency assertion would not allow a processor to stop, but require it to proceed with the content of the top-most compliant hierarchy level.

Stopping in a loud way seems less error-prone than silently relying on a convention to me. But I don't feel strongly about that, as long as the convention is clear.

>  require it to proceed with the content of the top-most compliant hierarchy level.

What does that mean, how do you determine which one is "compliant" if line text and concatenated words differ? @tboenig Can you clarify https://github.com/OCR-D/assets/issues/12#issuecomment-422020664?

bertsky:
> Stopping in a loud way seems less error-prone than silently relying on a convention to me.

Same here.

> > require it to proceed with the content of the top-most compliant hierarchy level.
> 
> What does that mean, how do you determine which one is "compliant" if line text and concatenated words differ?

Sorry, I mis-represented the rule stated by Matthias. I think he really meant using the lowest available hierarchy level. This may be difficult for the processor, though (it might require re-building the text for the needed level from a lower level).

page.md
```diff
@@ -36,6 +43,68 @@ The `comments` attribute of the `<pg:AlternativeImage>` attribute should be used
...
+the value `1`.
+
+## Consistency of text results on different levels
+
+Since text results can be defined on different levels and those levels can
+be nested, text results information is redundant. To avoid inconsistencies,
+the following assertions must be true:
+
+  * text of `<pg:Word>` must be equal to the text results of all `<pg:Glyph>`
+    contained within, concatenateb with empty string
```
kba: concatenateb -> concatenated

page.md
```diff
@@ -36,6 +43,68 @@ The `comments` attribute of the `<pg:AlternativeImage>` attribute should be used
...
+  * text of `<pg:TextLine>` must be equal to the text results of all
+    `<pg:Word>` contained  within, concatenated with a single space (`U+0020`).
+  * text of `<pg:TextRegion>` must be equal to the text results of all
+  * `<pg:TextLine>` contained within, concatenated with a newline (`U+000A`).
+
+**NOTE:** "Concatenation" means joining a list of strings with a separator, no
```
kba: Add a NOTE how to handle multiple textequivs by index.

page.md
```diff
@@ -36,6 +43,68 @@ The `comments` attribute of the `<pg:AlternativeImage>` attribute should be used
...
+separator is added to the start or end of the resulting string.
+
+If any of these assertions fails for a PAGE document, a processor must proceed
+with the text results at the lowest level provided.
+
+**Example:** `<pg:Word>` has text `Fool` but contains `<pg:Glyph>` whose
+text results, when concatenated, form the text `Foot`. The processor must proceed as if
+the `<pg:Word` had the text `Foot`.
```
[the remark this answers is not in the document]
kba: Different segmentations are an issue, c.f. https://github.com/OCR-D/spec/issues/72, but this proposal here is simpler.

First of all, it allows us to identify inconsistencies between the levels.

Secondly, if there are inconsistencies, it's always the lowest level that "wins".

IMHO, this should not happen at all and a processor should be expected to either produce consistent `TextEquiv` on all levels (processor changed words-> adapt lines, blocks) or delete those `TextEquiv` that cannot be kept consistent (processor changed words -> delete glyphs).

> Question: Why save a textequiv for each level if the textequivs are overwritten by the deeper level anyway?

Convenience I suppose. Since enforcing these consistency rules requires generating and comparing text anyway, we could probably prune the upper levels after processing and repopulate them before processing.

> What will happen if multiple textequivs exists? Must the indexes match?

Yes, that's a good point I'll add a note on that.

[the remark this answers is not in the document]
kba: > What will happen if multiple textequivs exists? Must the indexes match?

Updated, @VolkerHartmann please have a look at f94c9bb

bertsky:
> time there is a change

@VolkerHartmann Can you please elaborate on what you mean by that? I thought PAGE annotations are never changed (except perhaps for dev mode with our new rollback operation, where it would be overwritten). So far I was told that processors can only add new annotations.

> Add a NOTE how to handle multiple textequivs by index.

@kba I am quite surprised by the new consistency rules for multiple results in f94c9bb. In my understanding, higher levels _necessarily_ need more alternatives to account for the same information, because they are combinations of their sub-elements' alternatives. Even if the combination is not expanded in full, e.g. a `Word` does not get all possible `TextEquiv` from its constituent `Glyph`s `TextEquiv`s but only those which can be found in some dictionary or generated by some morphological rule, it would be overly restrictive to demand the same number of alternatives as its constituents. Also, the current formulation implies that all elements (not just the most ambiguous) need the same number of alternatives. Is that even realistic?

kba:
> I am quite surprised by the new consistency rules for multiple results in f94c9bb. In my understanding, higher levels necessarily need more alternatives to account for the same information, because they are combinations of their sub-elements' alternatives. Even if the combination is not expanded in full, e.g. a Word does not get all possible TextEquiv from its constituent Glyphs TextEquivs but only those which can be found in some dictionary or generated by some morphological rule, it would be overly restrictive to demand the same number of alternatives as its constituents. Also, the current formulation implies that all elements (not just the most ambiguous) need the same number of alternatives. Is that even realistic?

I'm struggling with that part as you noticed but wanted to propose something lest we forget this additional source of inconsistencies. I'd be happy about better/cleaner wording and less convoluted rules. "Same cardinality of multiple text equivs" throughout makes only sense when combining results, not for alternatives.

Would it be acceptable if consistency was restricted to the "first" text result? This would leave the processor the freedom to handle alternatives as it pleases and we wouldn't need to define it (and implement it, which I imagine to be quite hairy considering all the edge cases and different interpretations of what multiple text results may mean). Essentially, we'd only ensure consistency for the "canonical" text results.

bertsky:
> "Same cardinality of multiple text equivs" throughout makes only sense when combining results, not for alternatives.

If multiple `TextEquiv` results are not _alternative_ results, what else are they? I get the feeling I am missing some important in the whole discussion...

> Would it be acceptable if consistency was restricted to the "first" text result?

I'd recommend against that. That would shift the burden of checking consistency to the consuming processor again (at least for those processors that do consume multiple results / alternatives). 

Nevertheless, I completely agree something must be stated about this in the spec, and hopefully also enforced by the validator in core. 

I think that most of your original idea can be saved if you replace the notion of identical cardinality by subsumption: a level's `TextEquiv`s must each be subsumed in the concatenated combinations of its subordinate level's `TextEquiv`s. This is easy to check without expanding the full combination (which could be practically infeasible), for instance like this (untested):
```python
def validate_level(text, subs):
    if not subs:
        return True # end of recursion
    subtexts = subs[0].get_TextEquiv()
    for subtext in subtexts:
        prefix = subtext.get_Unicode()
        # for Word/TextLine/TextRegion: append space/newline delimiter to prefix here
        if text.startswith(prefix):
            if validate_level(text[len(prefix):], subs[1:]):
                return True # consistent
        pass # backtrack to next alternative
    return False # inconsistent
# e.g. at some Word element word:
texts = word.get_TextEquiv()
glyphs = word.get_Glyph()
for text in texts:
    if not validate_level(text.get_Unicode(), glyphs):
        raise Exception("word %s alternative '%s' (index %d) is inconsistent w.r.t. its constituent glyphs" % (word.get_id(), text.get_Unicode(), text.get_index())
```

page.md
```diff
@@ -8,19 +8,26 @@ representing a PAGE document](https://ocr-d.github.io/mets#media-type-for-page-x
...
 ## Original image as imageFilename
 
-The `imageFilename` attribute of the `<pg:Page>` MUST reference the original image and MUST NOT change between processing steps.
+The `imageFilename` attribute of the `<pg:Page>` MUST reference the original
+image and MUST NOT change between processing steps.
 
 ## AlternativeImage for derived images
 
-To encode images derived from the original image, the `<pc:AlternativeImage>` should be used. Its `filename` attribute should reference the URL of the derived image.
+To encode images derived from the original image, the `<pc:AlternativeImage>`
```
cneud: "should" or "MUST"?

page.md
```diff
@@ -36,6 +43,79 @@ The `comments` attribute of the `<pg:AlternativeImage>` attribute should be used
 
 ## AlternativeImage on sub-page level elements
 
-For the results of image processing that changes the positions of pixels (e.g. cropping, rotation, dewarping), `AlternativeImage` on page level and polygon of recognized zones is not enough to access the image section a region is based on since coordinates are always relative to the original image.
+For the results of image processing that changes the positions of pixels (e.g.
+cropping, rotation, dewarping), `AlternativeImage` on page level and polygon of
+recognized zones is not enough to access the image section a region is based on
```
cneud:
```suggestion
recognized zones is not sufficient for accessing the section of the image that a region is based on
```

page.md
```diff
@@ -36,6 +43,79 @@ The `comments` attribute of the `<pg:AlternativeImage>` attribute should be used
...
+since coordinates are always relative to the original image.
+
+For such use cases, `<pg:AlternativeImage>` may be used as a child of
+`<pg:TextRegion>`, `<pg:TextLine>`, `<pg:Word>` or `<pg:Glyph>`.
+
+## Attaching text recognition results to elements
+
+A PAGE document can attach recognized text to different typographical units of
```
cneud:
```suggestion
A PAGE document can attach recognized text to typographical units of
```

page.md
```diff
@@ -36,6 +43,79 @@ The `comments` attribute of the `<pg:AlternativeImage>` attribute should be used
...
+Leading and trailing whitespace (`U+0020`, `U+000A`) in the content of a
+`<pg:Unicode>` is not significant and must be removed from the string by
+processors.
+
+To encode an actual space character at the start or end of the content
+`<pg:Unicode>`, use a non-breaking space `U+00A0`.
+
+## Text recognition confidence
+
+The confidence of the correctness of the text recognition results in a
```
cneud:
```suggestion
The confidence score describing the assumed correctness of the text recognition results in a
```

page.md
```diff
@@ -36,6 +43,79 @@ The `comments` attribute of the `<pg:AlternativeImage>` attribute should be used
...
+`<pg:TextEquiv>` can be expressed in an attribute `@conf` as a float value
+between `0` and `1`, where `0` means "certainly wrong" and `1` means "certainly
+correct".
+
+## Attaching multiple text recognition results to elements
+
+Alternative text recognition results can be expressed by using multiple
+`<pg:TextEquiv>` wherever a single `<pg:TextEquiv>` would be allowed. When
+using mulitple `<pg:TextEquiv>`, they each must have an attribute `@index` with
```
cneud:
```suggestion
using multiple `<pg:TextEquiv>`, they each must have an attribute `@index` with
```

page.md
kba: MUST

kba: Ah, now I remember why it is `SHOULD` and not `MUST`: We do not use this mechanism at the moment. Rather we explicitly demand input file groups on call to a processor. So there is little benefit in setting it in the PAGE XML file.

If we demand it, we must implement it...

bertsky:
Still no ideas about the problem with multiple (alternative) results? No opinions on my proposal?

kba:
> Still no ideas about the problem with multiple (alternative) results? No opinions on my proposal?

IIUC @wrznr @cneud you prefer a "consistency checked for first alternative only" approach?

cneud:
Indeed I am currently leaning towards the "consistency checked for first alternative only" approach but this being a fairly tricky issue I want to make sure that I am not overlooking anything. I will utter my opinion by Tuesday next week at the latest.

cneud:
In lack of a better idea, I'd opt for the way currently proposed by @kba , i.e. we only concern ourselves with the consistency between the upper level and the top-ranked (index position = 1) alternative result on the lower level. Anything else would too quickly escalate into gazillions of combinations which will be impossible to keep consistent.

Even using subsumption instead as proposed [here](https://github.com/OCR-D/spec/pull/82#issuecomment-434858240), I cannot currently see how this will reduce the scope to keep track of (but will be happily enlightened). 

What this entails however, is that any alternatives present on a lower level that are not consistent with the upper level MUST have an index position greater than 1. 

To illustrate:
``format``
  &nbsp;&nbsp;``f``&nbsp;&nbsp;``f``
  &nbsp;&nbsp;``o``&nbsp;&nbsp;``o``
  &nbsp;&nbsp;``r``&nbsp;&nbsp;``r``
  &nbsp;&nbsp;``m``&nbsp;&nbsp;``r``
  &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;``n``

In this case, the ``n`` - since/while not included in the index = 1 variant - MUST have index > 1.

Needless to say, we will strive for a clear explanation (ideally illustrated by examples) asap. 

Last but not least, pinging @finkf again since this will likely also impact the work on [ocrd-postcorrection](https://github.com/cisocrgroup/ocrd-postcorrection) - and I believe this issue may have come up already earlier in the development of [PoCoTo](https://github.com/cisocrgroup/PoCoTo)?

bertsky:
> Even using subsumption instead as proposed [here](https://github.com/OCR-D/spec/pull/82#issuecomment-434858240), I cannot currently see how this will reduce the scope to keep track of (but will be happily enlightened).

We need to be more precise here: IINM _keeping track of_ in that context can only mean validating some element's TextEquiv against its constituent (i.e. lower) elements' TextEquivs, should such exist. There is no need to go into a full combination of **all** TextEquivs of all its sub-elements here (or with even larger scope, which then indeed would be gazillions): we already **know** the result text we are looking for. We can do that search by a simple back-tracking recursion, as illustrated, in a function that can be used for every level (it only needs to know what element delimiters the respective level requires). The complexity (and practically, stack depth) of that recursion grows with the number of sub-elements times their average ambiguity. Even in the _worst case_, one full TextLine of very short Words with lots of alternative readings, or equivalently one very long Word with lots of alternative Glyph readings, that problem will be small computationally IMO.

And that cost will be paid for regardless of whether we validate consistency in the workspace or in the consumer itself: Because any sensible processor that does take alternatives and that does consider multiple hierarchy levels (and thus necessitates consistency checking in the first place) will rely on alternatives subsumption. 

So to me this is actually a configurational issue: Workflows that have such processors would need that kind of validation at some point, but workflows without producers of ambiguity or without multi-level consumers would not. 

If my assessment is correct, always validating levels but merely validating index 1 is both too much and too little at the same time.

cneud:
@bertsky OK, that makes it clear, thanks for elaborating. 

However, when we discussed this again in the wider OCR-D coordination group, we felt that the back-tracking recursion approach would introduce considerable complexity in the code with regard to maintenance and testing compared to the added benefit and that we would thus prefer proceeding with the implementation of the "consistency checked for first alternative only" method instead to ensure at least a practical level of consistency.

We hope this is acceptable for you and do of course remain open to further discussions.

bertsky:
@cneud sure, no problem. After all, I can still do it in the processor. 

Hopefully, those consistency problems can already be avoided by design anyway when we extend PAGE with graph-based or matrix-based alternatives.

bertsky:
> In such a (inconsistent) case, what would you recommend the source of truth should be, the word or the glyphs? Your example seems particular to the word-glyph relation or do you expect TextLine and containing Words / TextRegion and containing TextLines to diverge because of language model/post-correction?

I guess it all depends on what you ask for: If I need Glyph level annotation and ask Tesseract to recognize one TextLine at a time, accessing results with `GetIterator` and `GetChoiceIterator`, it will give index 1 consistency on all 3 levels. But if I access results with `GetBestLSTMChoices` (or some future low-level routine), it will usually be inconsistent by the index 1 principle. Likewise if I ask to recognize each level separately (external layout analysis). 

A consumer that needs Glyphs would naturally like to ignore inconsistent higher levels, but a consumer that needs Words or TextLines would rather like the input repaired from the next-lower consistent (but not necessarily the lowest) level.

So maybe we should first look at the use-cases of _producing_ text annotation on multiple levels again (please cmiiw):
1. readability: providing a TextEquiv index 1 on the levels **higher** than current helps humans quickly navigate through PAGE-XML and comes without extra effort. Here index 1 consistency should be enforced.
2. avoid loosing granularity: providing a TextEquiv index 1 on the levels **lower** than current allows to also pass on all features which are available with higher resolution (in the input already or just now; like coordinates, font features etc), so each output is self-contained and no re-alignment of annotations becomes necessary. This case also comes with little extra cost, and index 1 consistency is appropriate here as well.
3. "one size fits all": providing TextEquivs (possibly with multiple indexes) on multiple levels can seem useful if one does not know what level consumers will expect, or if one knows the consumers to expect different levels. This can be expensive though, and index 1 consistency is overly restrictive here.

Case 3 really is a configuration issue and should probably dealt with differently (e.g. by calling the producer exactly once per required level; see #77). 

Coming back to our Tesseract example:
- `GetBestLSTMChoices` with case 1: To abide by the index 1 consistency principle, we should not use the result from `GetUTF8Text` on the higher levels – throwing away all lexicon, LM and heuristic knowledge going into it –, but instead only concatenate from each lower level, respectively. Or avoid higher level annotation for humans, or avoid that function completely.
- external layout, for example calling with a different model/language/engine on a low-confidence substring: we must ensure in the wrapper that higher levels are recombined consistently and lower levels use `GetIterator`.

Thus, probably I was wrong, and the "index 1 stays index 1" principle can be upheld sensibly. Even for processors using alternatives, if configured appropriately, the producer and consumer will agree on the hierarchy level they operate (provide/expect alternatives) on, and the other levels will merely contain one reading.

> > In case of failure, why does it require to use the lowest annotation level globally (on the document level), instead of the lower of the 2 levels in question, for that particular spot within in the document?
> 
> That was not the intended meaning. I had in mind an algorithm like this:

Oh, I see. So your intended meaning was element-local, next consistent level backup. But then the wording on L119-120 should IMO be changed to:
`If any of these assertions fail for an element, a processor must proceed
with the text results at the lower level.`

Regarding the actual code, how do you get a consistent (repaired) TextRegion text from the page element in the 4th line? You probably meant it this way:
```python
    if elem.isConsistent():
        text = elem.get_TextEquiv().get_Unicode()
    else:
        text = elem.concatenateConsistently()  # construct from lower level at this particular place in the document # probably needs to be recursive to find the next-highest lower level which is still consistent in itself
    # work with text
```

> You mean make consistency errors fatal?
> 
> > > As far as I understood @tboenig in this comment, failing the consistency assertion would not allow a processor to stop, but require it to proceed with the content of the top-most compliant hierarchy level.
> 
> Keep in mind that the origin of this proposal about consistency was to spot incorrect order of words in a line. But if strict consistency rules would restrict valid use cases too much, might it not be best to just scrap them?

You are absolutely right. If inconsistency is unavoidable in some use cases, and repairs are still possible, then consistency checking should detect this.

Issue #81, opened 2018-09-27
METS: Mapping by page must be by physical structmap


kba:
Currenly, we use the `GROUPID` of files to group different manifestations of the same file.

The more correct (and wider used) way is to create `mets:div` for every page in a `mets:structMap[TYPE="physical"]` and list `mets:ftpr` to all representations.

@tboenig @maria-federbusch Did I get that right?

wrznr:
@tboenig Push.

maria-federbusch:
You are right. The described way via mets:structMap[TYPE="physical"] is more correct.

PR #83, opened 2018-10-09
change page grouping according to #81

kba:
Must be by explicit physical structMap and file pointers, not implicitly via GROUPID

### DIFF
--- a/cli.md
+++ b/cli.md
@@ -36,8 +36,8 @@ File group(s) used as output.
 
 ### `-g, --group-id ID`
 
-The `ID` of the `mets:div[@TYPE='page']` that contains the pointers to files
-representing a page. Effectively, only those files in the [input
+The `mets:div[@TYPE='page']/@ID` that contains the pointers to files
+representing a page. Effectively, only those files in the [input file
 group](#-i---input-file-grp-grp) that are referenced in these
 `mets:div[@TYPE="page"]` will be processed.
 
--- a/cli.md
+++ b/cli.md
@@ -34,9 +34,12 @@ File group(s) used as output.
 
 ## Optional parameters
 
-### `-g, --group-id GROUPID`
+### `-g, --group-id ID`
 
-The `mets:file@GROUPID` to process. If no groupID is defined all files within the `inputGrp` will be processed. Repeatable and multiple IDs can be provided separated by comma.
+The `ID` of the `mets:div[@TYPE='page']` that contains the pointers to files
+representing a page. Effectively, only those files in the [input
+group](#-i---input-file-grp-grp) that are referenced in these
+`mets:div[@TYPE="page"]` will be processed.
 
 ### `-p, --parameter PARAM_JSON`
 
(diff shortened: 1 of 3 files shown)


mets.md
```diff
@@ -159,7 +170,7 @@ Always use URL. If it's a local file, prefix absolute path with `file://`.
 
 ```xml
 <mets:fileGrp USE="OCR-D-SEG-BLOCK">
-    <mets:file ID="OCR-D-SEG-PAGE_0001" GROUPID="OCR-D-IMG_0001" MIMETYPE="application/vnd.prima.page+xml">
+    <mets:file ID="OCR-D-SEG-PAGE_0001" MIMETYPE="application/vnd.prima.page+xml">
```
VolkerHartmann: Did we overlook that? Doesn't it have to be OCR-D-SEG-BLOCK_0001 instead of OCR-D-SEG-PAGE_0001 as file@ID?

mets.md
```diff
@@ -159,7 +170,7 @@ Always use URL. If it's a local file, prefix absolute path with `file://`.
...
         <mets:FLocat xmlns:xlink="http://www.w3.org/1999/xlink" LOCTYPE="URL" xlink:href="file:///path/to/workingDir/segmentation/block/page_0001.xml" />
```
VolkerHartmann: Shouldn't the subdir be SEG-BLOCK or SEG/BLOCK? But in the end it doesn't matter.

cli.md
```diff
@@ -34,9 +34,12 @@ File group(s) used as output.
 
 ## Optional parameters
 
-### `-g, --group-id GROUPID`
+### `-g, --group-id ID`
 
-The `mets:file@GROUPID` to process. If no groupID is defined all files within the `inputGrp` will be processed. Repeatable and multiple IDs can be provided separated by comma.
+The `ID` of the `mets:div[@TYPE='page']` that contains the pointers to files
```
VolkerHartmann: Alternatively: The `mets:div[@TYPE='page']@ID` that

cli.md
```diff
@@ -34,9 +34,12 @@ File group(s) used as output.
...
+representing a page. Effectively, only those files in the [input
+group](#-i---input-file-grp-grp) that are referenced in these
```
VolkerHartmann: Suggestion: Effectively, only those files in the [input file group] (#....)

cli.md
```diff
@@ -34,9 +34,12 @@ File group(s) used as output.
 
 ## Optional parameters
 
-### `-g, --group-id GROUPID`
+### `-g, --group-id ID`
```
VolkerHartmann: Suggestion: -g, --page-id ID
group-id is only understandable if you have the old definition in mind.
Unfortunately -p is already defined.

mets.md
kba: I think you're right.

mets.md
kba: Yeah, the point here was to have an arbirtrary file URL I think.

cli.md
kba: If so, then `mets:div[@TYPE='page']/@ID` so it's a proper XPath. No preference as long as it's consistent.

cli.md
kba: :+1:

cli.md
kba: It is clearer true, but we'd have to change a lot of code for it.

PR #91, opened 2018-10-26
ocrd-tool: allow definition of input/output file groups

kba:
@bertsky

### DIFF
--- a/ocrd_tool.md
+++ b/ocrd_tool.md
@@ -19,6 +19,20 @@ Optionally, workflow processors can be notified that this file is potentially
 large and static (e.g. a fixed dataset or a precomputed model) and should be cached
 indefinitely after download by setting the `cacheable` property to `true`.
 
+## Input / Output file groups
+
+Tools should define the names of expected input file groups as a list of `USE`
+attributes of `mets:fileGrp` elements. If more than one file group is expected, this
+should be explained in the description of the tool.
+
+The output file group of the tool should be provided as a string, the `USE`
+attribute of the `mets:fileGrp` where the tool writes its output files by default.
+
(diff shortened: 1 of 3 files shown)


bertsky:
Splendid!

ocrd_tool.md
```diff
@@ -59,6 +73,16 @@ properties:
           executable:
             description: The name of the CLI executable in $PATH
             type: string
+          input_file_grp:
+            description: Input fileGrp@USE this tool expects by default
+            type: array
+            items:
+              type: string
+              pattern: '^OCR-D-[A-Z0-9-]+$'
```
VolkerHartmann: Typo?: [A-Z0-9\\-]

ocrd_tool.md
```diff
@@ -59,6 +73,16 @@ properties:
...
+          output_file_grp:
+            description: Output fileGrp@USE this tool produces by default
+            type: string
+            pattern: '^OCR-D-[A-Z0-9-]+$'
```
VolkerHartmann: Typo?: [A-Z0-9\\-]
The backslash inside the square bracket has to be escaped by an additional backslash.

ocrd_tool.md
kba: dash at the end of a character group is not interpreted as a meta char, c.f. https://regex101.com/r/4d8P1b/1.

kba:
Should probably be even required, otherwise nobody will use it and workflow engine cannot rely on it.

ocrd_tool.md
VolkerHartmann: Ok, I didn't know that yet.

PR #94, opened 2018-11-13
First draft for logging and provenance.

VolkerHartmann:


provenance.md
```diff
@@ -0,0 +1,24 @@
...
+
+### Input/Output
+Input should be the input files referenced via FileIDs. Optionally the parameter file.
+Output should be the generated files referenced via FileIDs.
+
+### Ingest/Access Workspace to/from OCR-D Repositorium
+The linking of inputs and outputs should be done via the FileID as long as the files are not archived. 
+If the files are in repository, the FileIDs should be replaced by the unique URL.
+When exporting a workspace, the references should be reset to the FileID again.
+During the ingest, the entire provenance must be stored in a large graph to make the provenance searchable.
```
cneud:
```suggestion
During the ingest, the entire provenance must be stored in a graph to make the provenance searchable.
```

provenance.md
```diff
@@ -0,0 +1,24 @@
...
+
+## METS
+### File Group
+File group holding provenance have to start with prefix "PROV-"
```
cneud:
```suggestion
File groups holding provenance have to start with prefix "PROV-"
```

logging.md
```diff
@@ -0,0 +1,34 @@
+# Conventions for LOGGING
+[See issues](https://github.com/OCR-D/spec/issues/11)
+All logging may be stored in files. 
+If so, they must be given in Provenance as a result. 
+
+## Log Levels
+(Is there any definition we could link to)
```
cneud:
```suggestion
(Is there any definition we could link to)
```

cneud: Unfortunately I am not aware of any...[RFC 5424](https://tools.ietf.org/html/rfc5424) is not super verbose on this and the best explanation I could find is from [Stackoverflow](https://stackoverflow.com/a/5278006/1919465).

logging.md
```diff
@@ -0,0 +1,34 @@
...
+Note that logging within loops can dramatically affect performance.
+### DEBUG
+This level is used if parameters and or the status of an algorithm/method should be logged. 
+### INFO
+Log information about used settings on application level.
+### WARN
+This level is used to log information about missing/wrong configurations which may lead to errors.
+### ERROR
+Log all events that produce no or a wrong result. 
+It is useful to provide so much information that could provide information about the cause.
```
cneud: This line isn't really clear to me?

logging.md
```diff
@@ -0,0 +1,34 @@
...
+
+
+## Format
+The format of the logging output have to formatted like this:
```
cneud:
```suggestion
The format of the logging output has to be formatted like this:
```

logging.md
```diff
@@ -0,0 +1,34 @@
...
+TIMESTAMP LEVEL LOGGERNAME - MESSAGE
+***Example:***
+08:03:40.017 WARN edu.kit.ocrd.MyTestClass - A warn message
+:warning: 
+Since only the time stamp is logged, the log should be written to files that change daily.
+
+## METS
+### File Group
+File group holding logging have to start with prefix "LOG-"
```
cneud:
```suggestion
File groups holding logging have to start with prefix "LOG-"
```

logging.md
```diff
@@ -0,0 +1,34 @@
+# Conventions for LOGGING
+[See issues](https://github.com/OCR-D/spec/issues/11)
+All logging may be stored in files.
```
kba: It should be made clear that this is a different spec for different target audience than e.g. the [docs on the CLI](https://ocr-d.github.io/cli), which states

> Data printed to `STDERR` and `STDOUT` is captured linewise and stored as log data.

Maybe an introductory section, explaining the purpose of the document, who should read it and which are the essential requirements to implement.

logging.md
```diff
@@ -0,0 +1,34 @@
+# Conventions for LOGGING
+[See issues](https://github.com/OCR-D/spec/issues/11)
+All logging may be stored in files. 
+If so, they must be given in Provenance as a result.
```
kba: Provenance capitalized sounds kind-of biblical :) Seriously though, I'm missing context here, I don't understand what `if` is related to, what `being given in provenance` means and what a `result` is here.

logging.md
```diff
@@ -0,0 +1,34 @@
+# Conventions for LOGGING
+[See issues](https://github.com/OCR-D/spec/issues/11)
+All logging may be stored in files. 
+If so, they must be given in Provenance as a result. 
+
+## Log Levels
+A more detailed description will be found [here](https://stackoverflow.com/questions/2031163/when-to-use-the-different-log-levels/5278006#5278006)
```
kba: Since this is actually pertinent to developers IMHO it should be added to https://ocr-d.github.io/cli#logging.

The `core` implementation doesn't distinguish between TRACE and DEBUG.

logging.md
```diff
@@ -0,0 +1,34 @@
...
+### INFO
+Log information about used settings on application level.
+### WARN
+This level is used to log information about missing/wrong configurations which may lead to errors.
+### ERROR
+Log all events that produce no or a wrong result. 
+It is useful to output all the information that helps to determine the cause without the need for further investigation.
+
+
+## Format
```
kba: Already defined in https://ocr-d.github.io/cli#logging

logging.md
```diff
@@ -0,0 +1,34 @@
...
+The format of the logging output has to be formatted like this:
+TIMESTAMP LEVEL LOGGERNAME - MESSAGE
+***Example:***
+08:03:40.017 WARN edu.kit.ocrd.MyTestClass - A warn message
+:warning: 
+Since only the time stamp is logged, the log should be written to files that change daily.
```
kba: IIUC, you want logging to be the outcome of a job/task/process right (because of the bit about provenance at the top). That would not be a continuous runnign process where log rotation was important. Or do you want to concatenate the logging files of processes into a big file?

logging.md
```diff
@@ -0,0 +1,34 @@
...
+
+## METS
+### File Group
+File groups holding logging have to start with prefix "LOG-"
```
kba: So in terms of BagIt, you see log data as part of the payload? Provenance metadata, parametrization etc. also part of the payload? How to map between provenance, log data, processes and files?

provenance.md
```diff
@@ -0,0 +1,24 @@
+# Provenance
+[See issues](https://github.com/OCR-D/spec/issues/11)
+All provenance have to be stored in files. 
+There's provenance at the page level, but there's also provenance at the document level.
+Because these files can become very large, they are also stored as references in the fileGrp section.
+
+## Format
+Not defined yet
+
+### Input/Output
```
kba: Since a step in the workflow will be a callout to a tool, I'd suggest basing the input on the concrete call, with defaults taken from the ocrd-tool.json.

provenance.md
```diff
@@ -0,0 +1,24 @@
...
+
+### Ingest/Access Workspace to/from OCR-D Repositorium
+The linking of inputs and outputs should be done via the FileID as long as the files are not archived. 
+If the files are in repository, the FileIDs should be replaced by the unique URL.
+When exporting a workspace, the references should be reset to the FileID again.
+During the ingest, the entire provenance must be stored in a graph to make the provenance searchable.
+
+## METS
+### File Group
+File groups holding provenance have to start with prefix "PROV-"
```
kba: See above, I'm not convinced whether provenance and logging data should be part of METS at all.

I think it would be sufficient to record `mets:agent` for the workflow that touched a METS with a URL to a workflow engine run which then allows further introspection.

logging.md
VolkerHartmann: I'll write something.

logging.md
VolkerHartmann: True. Trace is only for completeness and should not be used in the workflow.

logging.md
VolkerHartmann: This is a clumsy phrase. The output should never be collected in a large file. The statement should be that the output can always be assigned to a work step.

logging.md
VolkerHartmann: At least the provenance should be part of the payload. The provenance refers to the input/output files (via ID?), which can then be used for assignment. All provenances must then be combined to form a graph.

provenance.md
kba: Yes I agree, it should be tracked but should it be part of the payload? Or wouldn't it make more sense to have

* in the METS only a reference to the URL representing the workflow execution(s) that modified this file with OCR-D tools. Using a mets:agent for example.

* Make that URL dereferenceable for provenance and logging metadata

* Bundle the data in the container (bag) but not the METS file itself?

logging.md
kba: > At least the provenance should be part of the payload.

Still don't understand why. Provenance is by definition metadata and not about the thing but about the process that made the thing. Why should this be different from logging data e.g.?

My main problem here is that METS consists (for our purposes!) of files representing pages in file groups representing different manifestations of these pages. The file groups are the output of processes but they don't reflect processes directly.

Provenance is based on processes which can have arbitrary input/output. With special filegroups that don't reflect pages anymore, this consistency breaks. If provenance was to be tracked in METS (which I would not advise), I see these options:

1) Have a file group `X-PROV` for every file group `X` which a `mets:file` for every `mets:file` in `X` that points to the URL of the process that generated that file
2) Add a custom attribute or (re-/ab-)use an existing attritute on the file itself
3) Have some sort of mapping structure that maps XML IDs to the URL of the process that created them

The last one would at least be general enough to also handle cases like "which process created the table of contents". But it will still be a convention not a standard, extend the size of METS forever and have little use for the institutions (who have their own digitization workflow engines which have their own naming schemes and conventions for provenance).

@cneud @wrznr

provenance.md
VolkerHartmann: I always struggle with the term metadata, because PAGE XML is also called metadata in other use cases. 
We may also define a tag directory metadata which contains a provenance and log directory holding all collected metadata. But it is difficult to refer to it from the METS file because it is a parent directory.
For provenance metadata we may reference the input/output files via their ID inside the METS file. The log metadata must be referenced using a job ID that should be part of the provenance metadata.

logging.md
VolkerHartmann: Provenance is important because it is the only way to reproduce data, but reproducibility is an essential characteristic of scientific activity.
By (our) definition, all files contained in the bagit container must also be referenced in METS. As a result, the provenance must also be referenced in METS if it is part of the payload.
Optionally, we can define it as metadata that should be stored in a tag directory of the bagit container, but then the coupling is relatively loose.  (See comment below)
Also the bagit import/export to/from workspace will become more difficult.

cneud:
@kba Can we merge this for now and leave the discussion of further extensions for provenance to when @VolkerHartmann returns?

kba:
> Can we merge this for now

The provenance document is useful for discussion, though it's probably good to mark it as a draft. The part about logging seems IMHO largely redundant to the CLI docs.

Compromise: Add a disclaimer ("This document is still a draft.") to the files and merge it?

We should continue the discussion once @VolkerHartmann is back.

cneud:
Thx, I'll merge and mark as ``Draft``.

PR #97, opened 2018-11-13
Allow a number of files in the bag root dir

kba:
For ground truth and test assets.

Not an actionable bagit profile at the moment though, c.f. https://github.com/bagit-profiles/bagit-profiles/pull/14.


ocrd_zip.md
```diff
@@ -120,6 +120,18 @@ Wrong:
 All files except `mets.xml` itself that are contained in `data` directory must
 be referenced in a `mets:file/mets:Flocat` in the `mets.xml`.
 
+## Optional metadata about the payload
+
+In addition to the actual data files in `/data`, the following metadata files
+are allowed to be present in the root of the bag:
+
+* `README.md`: An extended, human-readable description of the dataset in the Markdown syntax
```
VolkerHartmann: The README.md should be part of the payload, right?
It would be good, of course, if we always had this.

ocrd_zip.md
```diff
@@ -120,6 +120,18 @@ Wrong:
...
+* `Makefile`: A GNU make build file to reproduce the data in `/data`.
+* `build.sh`: A bash script to reproduce the data in `/data`.
```
VolkerHartmann: Ideally for reproducibility, provenance metadata should be used.

ocrd_zip.md
kba: No, not part of the payload, only meta-information, such as description of test assets. In most (automated) cases, this rarely makes sense.

I like the rule to only have as payload what is referenced in METS.

ocrd_zip.md
kba: Provenance metadata is not always available, such as for ground truth, test data, example sets etc.

ocrd_zip.md
VolkerHartmann: I understand what you mean, but I'm a bit ambivalent. If we don't store this information in the payload, it might get 'lost'. And if this information exists, I'd find it useful if it were there.
However, if it is only used in test data, I agree with you.

kba: As you say, in production, there should be provenance data to carry this information.

PR #99, opened 2018-11-16
OCRD-ZIP clarify that relative paths are prefered over file URL

kba:
It's more consistent to have the restrictions on the payload be always relative to the `/data` directory in a bag.

It is also closer to reality and easy to detect whether a string is a relative path or a URL. It allows working with standard tools (without removing `file://` prefixes, prefixing `/data` or any using absolute paths).

The implementation does it like that, this just up the definitions.

### DIFF
--- a/ocrd_zip.md
+++ b/ocrd_zip.md
@@ -49,7 +49,7 @@ the full definition](#appendix-a)):
   * [`Ocrd-Identifier`](#ocrd-identifier): A globally unique identifier for this bag
   * [`Ocrd-Base-Version-Checksum`](#ocrd-base-version-checksum): Checksum of the version this bag is based on
 * `bag-info.txt` MAY additionally contain these tags:
-  * [`Ocrd-Mets`](#ocrd-mets): Alternative path to the mets.xml file, relative to `/data`, if its path IS NOT `mets.xml`
+  * [`Ocrd-Mets`]: Alternative path to the mets.xml file, relative to `/data`, if its path IS NOT `mets.xml`
   * [`Ocrd-Manifestation-Depth`](#ocrd-manifestation-depth): Whether all URL are dereferenced as files or only some
 
 ### `BagIt-Profile-Identifier`
@@ -151,17 +151,15 @@ To pack a workspace to OCRD-ZIP:
   * Replace the URL of `f` with the path relative to `/data` (SHOULD be `<USE>/<ID>`) in
     * all `mets:FLocat` of the METS
     * all other files in the workspace, esp. PAGE-XML
-* Write out the changed METS to `TMP/data/mets.xml` (or another location in `TMP/data/`, specified by [`Ocrd-Mets`](#ocrd-mets))
+* Write out the changed METS to `TMP/data/mets.xml`
 * Package `TMP` as a BagIt bag
 
 ### Unpacking OCRD-ZIP to a workspace
 
 * Unzip OCRD-ZIP `z` to a folder `TMP`
-* Let `M` be `TMP/data/mets.xml` (or another location in `TMP/data/`, specified by [`Ocrd-Mets`](#ocrd-mets))
-* Foreach file `f` in `M`:
-  * Strip `file://` from the beginning of the `xlink:href` of `f`
-  * If it is not a file path (begins with `http://` or `https://`):
-    continue
+* If the value `M` of [`Ocrd-Mets`] is different from `mets.xml`:
+  * Rename `TMP/data/mets.xml` to `TMP/data/` + `M`
+* Move `TMP/data` to an appropriate location to use as a workspace
 
 ## Appendix A - BagIt profile definition
 
@@ -209,3 +207,5 @@ Accept-BagIt-Version:
 Proposed media type of OCRD-ZIP: `application/vnd.ocrd+zip`
 
 Proposed extension: `.ocrd.zip`
+
+[`Ocrd-Mets`]: #ocrd-mets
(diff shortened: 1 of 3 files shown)


ocrd_zip.md
```diff
@@ -127,25 +130,26 @@ be referenced in a `mets:file/mets:Flocat` in the `mets.xml`.
...
 ### Unpacking OCRD-ZIP to a workspace
 
 * Unzip OCRD-ZIP `z` to a folder `TMP`
 * Foreach file `f` in `TMP/data/mets.xml`:
-  * If it is not a `file://`-URL, continue
-  * Replace the URL of `f` with `file://<ABSPATH>`, where `<ABSPATH>` is the absolute path to `f`, in
+  * If it is not a `file://`-URL and not a file path,
+    continue
+  * Replace the URL of `f` with `<ABSPATH>`, where `<ABSPATH>` is the absolute path to `f`, in
     * `TMP/data/mets.xml`
```
VolkerHartmann: or the file specified by [`Ocrd-Mets`](#ocrd-mets)

ocrd_zip.md
```diff
@@ -127,25 +130,26 @@ be referenced in a `mets:file/mets:Flocat` in the `mets.xml`.
...
+* Foreach `mets:file` `f` in the source METS:
+  * If it is not a `file://`-URL and not a file path
+    * If `Ocrd-Manifestation-Depth` is `partial`,
       continue
   * Download/Copy the file to a location within `TMP/data`. The structure SHOULD be `<USE>/<ID>` where
     * `<USE>` is the `USE` attribute of the parent `mets:fileGrp`
     * `<ID>` is the `ID` attribute of the `mets:file`
-  * Replace the URL of `f` with `file:///data/<USE>/<ID>` in
-    * all `mets:FLocat` of `TMP/data/mets.xml`
+  * Replace the URL of `f` with the path relative to `/data` (SHOULD be `<USE>/<ID>`) in
```
VolkerHartmann: But this is in contrast to what is written above.
> Right:
>  * `file:///data/foo.xml`

ocrd_zip.md
```diff
@@ -127,25 +130,26 @@ be referenced in a `mets:file/mets:Flocat` in the `mets.xml`.
 To pack a workspace to OCRD-ZIP:
```
VolkerHartmann: If only absolute paths are allowed and all files in the data directory must be referenced in METS, the workspace actually consists only of the METS file.

ocrd_zip.md
```diff
@@ -127,25 +130,26 @@ be referenced in a `mets:file/mets:Flocat` in the `mets.xml`.
...
       continue
   * Download/Copy the file to a location within `TMP/data`. The structure SHOULD be `<USE>/<ID>` where
     * `<USE>` is the `USE` attribute of the parent `mets:fileGrp`
     * `<ID>` is the `ID` attribute of the `mets:file`
-  * Replace the URL of `f` with `file:///data/<USE>/<ID>` in
-    * all `mets:FLocat` of `TMP/data/mets.xml`
+  * Replace the URL of `f` with the path relative to `/data` (SHOULD be `<USE>/<ID>`) in
+    * all `mets:FLocat` of the METS
     * all other files in the workspace, esp. PAGE-XML
+* Write out the changed METS to `TMP/data/mets.xml` (or another location in `TMP/data/`, specified by [`Ocrd-Mets`](#ocrd-mets)
```
VolkerHartmann: Is the bagit container now defined as workspace?
The 'source METS file' is the only METS file that can be contained in the Bagit container, as all other METS files that may be present are not referenced in the 'source METS file'.
If so there is no need to rename it.

ocrd_zip.md
```diff
@@ -127,25 +130,26 @@ be referenced in a `mets:file/mets:Flocat` in the `mets.xml`.
...
 
 ### Unpacking OCRD-ZIP to a workspace
 
 * Unzip OCRD-ZIP `z` to a folder `TMP`
 * Foreach file `f` in `TMP/data/mets.xml`:
-  * If it is not a `file://`-URL, continue
-  * Replace the URL of `f` with `file://<ABSPATH>`, where `<ABSPATH>` is the absolute path to `f`, in
+  * If it is not a `file://`-URL and not a file path,
+    continue
+  * Replace the URL of `f` with `<ABSPATH>`, where `<ABSPATH>` is the absolute path to `f`, in
```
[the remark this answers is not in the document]
kba: I want `/data` to be effectively a workspace when unzipped. My takeaway from the feedback to my `url-sources.txt` proposal was that we should just allow relative paths.

At the moment, the `ocrd zip spill` implementation won't modify any `mets:FLocat` because they are either relative filenames at this point or HTTP-URL.

Get rid of the absolute paths for good?

ocrd_zip.md
kba: > Is the bagit container now defined as workspace?

`/data` should be also useable as a workspace.

> If so there is no need to rename it.

In real life, the METS filename will usually be meaningful, like an ID of the work or such. It seems sensible to offer a way to preserve that.

But we could handle that transparently and consistenly name that file `data/mets.xml` within the OCRD-ZIP and only rename it on spilling the bag. Would that work for you?

ocrd_zip.md
kba: That was the original idea. But it's a hassle in reality, hence the workspace concept which combines METS with downloaded files.

ocrd_zip.md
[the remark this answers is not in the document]
kba: Relative files allowed in general METS too, now, see updated commits.

kba:
@VolkerHartmann thanks for the feedback, I tried to make the new relative-files-allowed policy more consistent.

ocrd_zip.md
VolkerHartmann: > It seems sensible to offer a way to preserve that.

I think so, too.
If there is only one METS file inside the bagit container how do we avoid duplicates in file names.
If there are multiple METS files inside a workspace/directory, which one is the 'right' one? This information gets lost when the context (bagit container or workflow) gets lost.

PR #101, opened 2018-12-11
Replace GROUPID mechanism, --group-id with --page-id

kba:


wrznr:
It should be on our long-term ToDo list.

kba:
> Eventually we should stay with `G` for paGe or use `T` for parameTer.

`-G` is used for the file group in `workspace add` though (also differs in case only). I see how using both uppercase and lowercase single params can be confusing but it is common practice in CLI and when in doubt, one can always use the long-form (e.g. `--page-id`).

I'll change it to `-g/--page-id` then.

PR #103, opened 2018-12-18
additional consistency level "lax"

kba:
Otherwise, punctuation encoded as `Word` are a nightmare of false positives and/or byzantine concatenation rules.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ Versioned according to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+Added:
+
+  * Consistency check level 'lax'
+
 Fixed:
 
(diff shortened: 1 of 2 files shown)


bertsky:
Wait, didn't we agree in OCR-D/assets#13 that punctuation is to be encoded as a separate `Word` if and only if it is preceded by _and_ followed by whitespace (as for example with spaced dash for parenthesis or pause)? 

I am lost, please help me out!

kba:
Yes, that would be best if punctuation is in a separate word iff it is whitespace separated.

Alas, in reality this isn't always the case.

A `lax` check would be a last resort to spot mixed up order of elements.

`strict` (the default) will raise punctuation-as-words as the errors they are.

bertsky:
Now I got it. Thanks!

PR #105, opened 2019-01-29
add 1st draft line GT/training specs

kba:


gt-profile.yml
```diff
@@ -0,0 +1,106 @@
...
+    default: '.gt.txt'
+  Gt-Transcription-Media-Type:
+    required: false
+    default: 'text/plain'
+  Gt-Transcription-Directory:
+    required: false
+    default: 'text'
+  Gt-Transcription-Normalization:
+    required: false
+    default: 'NFKC'
```
wrznr: Not sure if we should have a default here. I'd opt for `required` without default. Too much depends on a correct setting.

wrznr: In addition, the specification should require "utf-8" encoding for the GT text somewhere to make those normalizations work, right?

gt-profile.yml
```diff
@@ -0,0 +1,106 @@
...
+      # img and transcription in subfolders Gt-Bitonal-Image-Directory and Gt-Transcription-Directory of Gt-Directory
+      - 'subfolders'
+      # img and transcription in subfolders Gt-Bitonal-Image-Directory and Gt-Transcription-Directory in the same dir below Gt-Directory
+      - 'subfolders-nested'
+Manifests-Required: ['sha512']
+Tag-Manifests-Required: []
+Tag-Files-Required: []
+Tag-Files-Allowed:
+  - README.md
+Allow-Fetch.txt: false
```
wrznr: No idea what this is supposed to do.

gt-spec.md
```diff
@@ -0,0 +1,196 @@
...
+	* [Gt-Directory](#gt-directory)
+	* [Gt-Directory-Structure](#gt-directory-structure)
+* [Line metadata](#line-metadata)
+
+<!-- END-MARKDOWN-TOC -->
+
+## Rationale
+
+Recent OCR (optical character recognition) engines are not actually
+character-based anymore but on neural networks that operate on lines. These
```
wrznr: *but **use** neural*

gt-spec.md
```diff
@@ -0,0 +1,196 @@
...
+
+### Gt-Transcription-Normalization
+
+Unicode normalization level. One of `NFC`, `NFKC`, `NFD` or `NFKC`. Default: `NFKC`.
+
+![Illustration unicode normalization](http://unicode.org/reports/tr15/images/UAX15-NormFig6.jpg)
+
+### Gt-Grayscale-Image-Extension
+
+Extension of the grayscale image files. Default: `.png`.
```
wrznr: It's `*.nrm.png` above.

gt-spec.md
```diff
@@ -0,0 +1,196 @@
...
+
+Media type of the grayscale image files. Default: `image/png`.
+
+### Gt-Grayscale-Image-Directory
+
+Name of the subfolder containing grayscale images if [`Gt-Directory-Structure`] is `subfolders` or `subfolders-nested`. Default: `grayscale`.
+
+### Gt-Color-Image-Extension
+
+Extension of the color image files. Default: `.png`.
```
wrznr: It's `.color.png` above.

gt-spec.md
```diff
@@ -0,0 +1,196 @@
...
+
+Media type of the color image files. Default: `image/png`.
+
+### Gt-Color-Image-Directory
+
+Name of the subfolder containing color images if [`Gt-Directory-Structure`] is `subfolders` or `subfolders-nested`. Default: `img`.
+
+### Gt-Bitonal-Image-Extension
+
+Extension of the bitonal image files. Default: `.png`.
```
wrznr: It's `.bin.png` above.

gt-spec.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Line metadata
+
+In addition to the bag-wide metadata defined by the [BagIt profile], metadata
+can be saved per line to preserve the provenance of every single line.
+
+Line metadata can be encoded in JSON or YAML (depending on
+[`Gt-Line-Metadata-Extension`] and [`Gt-Line-Metadata-Media-Type`]).
+
+Line metadata mustt adhere to this JSON schema:
```
wrznr: mustt → must

training-schema.yml
```diff
@@ -0,0 +1,44 @@
...
+    type: string
+    default: '*'
+  outputModelFormat:
+    description: The output format of the model. Note that individual engines only support a single one or a subset of formats.
+    enum:
+      - application/vnd.ocrd.pronn  # kraken < 1.0
+      - application/vnd.ocrd.clstm  # ocropy-lpred, clstm, kraken<1.0
+      - application/vnd.ocrd.coreml # kraken >= 1.0
+      - application/vnd.ocrd.pyrnn  # ocropy-rpred
+      - application/vnd.ocrd.tf+zip # calamari, zipped tensorflow data
```
wrznr: This enum does not contain a tesseract-specific format (at least not in the comments).

gt-profile.yml
```diff
@@ -0,0 +1,106 @@
...
+  Gt-Bitonal-Image-Extension:
+    required: false
+    default: '.bin.png'
+  Gt-Bitonal-Image-Media-Type:
+    required: false
+    default: 'image/png'
+    values:
+      - 'image/png'
+      - 'image/tiff'
+      - 'image/jpeg'
```
mittagessen: JPEGs don't work for bitonal images as they are interpolated and always contain intermediate values. There might be some way to re-binarize them "losslessly" but it isn't straightforward.

gt-profile.yml
mittagessen: If requiring a value I'd add a custom one as there might be some sets that aren't normalized for whatever reasons.

gt-profile.yml
kba: Forbid incomplete bags, basically. Only the info in the `BagIt-Profile-Info` section and those properties beginning with `Gt-*` are of concern here.

training-schema.yml
kba: Right. How about `application/vnd.ocrd.tesseract4`?

kba:
> I strongly recommend the introduction of a third GT subset devel.

@wrznr Can you elaborate?

training-schema.yml
wrznr: Fine.

wrznr:
@kba 
> Can you elaborate?

Most training procedures allow for the application of three different sets of GT: `train`, `eval` and `devel`. While the purpose for the first two is supposedly clear, the latter is used during training for parameter fixing and error estimation. E.g. `ocropus-rtrain` has the parameter `--tests` for this purpose. Note that  	strictly speaking you may not abuse your evaluation data as development data.

wrznr:
@kba Cf. https://en.wikipedia.org/wiki/Training,_validation,_and_test_sets

cneud:
@wrznr So far I mainly applied [k-fold_cross-validation](https://en.wikipedia.org/wiki/Cross-validation_(statistics)#k-fold_cross-validation), would you still see added benefits over this by partitioning into three sets?

model-evaluation-schema.yml
```diff
@@ -0,0 +1,39 @@
+$id: https://ocr-d.github.io/schemas/v1/model-evaluation-schema.json
+type: object
+required:
+  - engineName
+  - engineVersion
+  - groundTruthBag
+  - model
+properties:
+  engineName:
```
kba: delete .Derived from model

model-evaluation-schema.yml
```diff
@@ -0,0 +1,39 @@
...
+    type: string
+    enum:
+      - ocropus
+      - kraken
+      - tesseract
+      - calamari
+  engineVersion:
```
kba: delete .Derived from model

gt-profile.yml
```diff
@@ -0,0 +1,115 @@
...
+      - NFKC
+      - non-normalized
+  Gt-Color-Image-Extension:
+    required: false
+    default: '.color.png'
+  Gt-Color-Image-Media-Type:
+    required: false
+    default: 'image/png'
+    values:
+      - 'image/png'
```
tboenig: Would a differentiation between Tiff compressed or JPEG2000 make more sense?

gt-profile.yml
```diff
@@ -0,0 +1,115 @@
+BagIt-Profile-Info:
+  BagIt-Profile-Identifier: https://ocr-d.github.io/gt-profile.json
+  BagIt-Profile-Version: '1.2.0'
+  Source-Organization: OCR-D
```
tboenig: What about information about the origin of the digitized lines?
- minimal bibliographic record based on DC?
- and artificially generated lines (+ degeneration)
- what about the degeneration algorithm?

I think that comment may be in the wrong place here. It should probably be placed in this place ## Line metadata##.

gt-profile.yml
```diff
@@ -0,0 +1,115 @@
...
+    required: false
+    default: 'img'
+  Gt-Grayscale-Image-Extension:
+    required: false
+    default: '.nrm.png'
+  Gt-Grayscale-Image-Media-Type:
+    required: false
+    default: 'image/png'
+    values:
+      - 'image/png'
```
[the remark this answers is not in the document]
kba: You mean additionally allow `image/jp2`? Do engines allow JPEG2000 input for training?

gt-profile.yml
kba: See https://github.com/OCR-D/spec/pull/105/files/6827085d051e945062203b82ef921e54025cfbda#diff-ee256e83a17cfe309565c88ab376091a That is the definition of what's currently supposed to be in there. Bibliographic metadata would be in the METS referred to by `metsUrl`. How to encode provenance on a line-level I am not sure though. @VolkerHartmann?

cneud:
@wrznr Do your remaining ``@wrznr requested changes`` relate to [this comment](https://github.com/OCR-D/spec/pull/105#pullrequestreview-197476416) only or is there other stuff that needs changing (for the time being)?

kba:
@Doreenruirui's work on okralact has diverged significantly from these specs. It makes little sense to publish these specs with the only implementation implementing it differently.

@doreenruirui can you compare your schemas and documentation with this so we can integrate that part of okralact into the specs?

Doreenruirui:
> @Doreenruirui's work on okralact has diverged significantly from these specs. It makes little sense to publish these specs with the only implementation implementing it differently.
> 
> @Doreenruirui can you compare your schemas and documentation with this so we can integrate that part of okralact into the specs?

@kba I am  sorry that I am not very familiar with github. Can you point me to the document I should compare with my schemas?

kba:
@Doreenruirui We're discussing these changes/new files: https://github.com/OCR-D/spec/pull/105/files.

In particular I would like to harmonize the proposal here (https://github.com/OCR-D/spec/pull/105/files?file-filters%5B%5D=.md#diff-2ae93b1f468c44b9f7e195133a0fb539) of using BagIt for the line GT with your approach in okralact wrt to input format.

Also interesting would be to compare okralact's engine schemas with the schema proposed here https://github.com/OCR-D/spec/pull/105/files?file-filters%5B%5D=.md&file-filters%5B%5D=.yml#diff-690d5874f98dfbd6737bc0168b6084d8 and https://github.com/OCR-D/spec/pull/105/files?file-filters%5B%5D=.md&file-filters%5B%5D=.yml#diff-a1f62fd4dd219fc5c5d5f0ccb419c88b

PR #106, opened 2019-01-30
Fetch partial

VolkerHartmann:
Unfortunately all files have to be fetched to validate bag due to specification of kunze!

Only the last (two?) commits please.

### DIFF
--- a/ocrd_zip.md
+++ b/ocrd_zip.md
@@ -133,6 +133,11 @@ Invalid `mets:FLocat/@xlink:href` in `/tmp/foo/ws1/data/mets.xml`:
 All files except `mets.xml` itself that are contained in `data` directory must
 be referenced in a `mets:file/mets:Flocat` in the `mets.xml`.
 
+### When in METS and not in data
+
+Due to partial OCRD-ZIP not all files may be part of the payload. If so they have to be 
+mentioned in fetch.txt and in all payload manifest files. 
+
 ## Optional metadata about the payload
 
 In addition to the actual data files in `/data`, the following metadata files


ocrd_zip.md
```diff
@@ -133,6 +133,11 @@ Invalid `mets:FLocat/@xlink:href` in `/tmp/foo/ws1/data/mets.xml`:
 All files except `mets.xml` itself that are contained in `data` directory must
 be referenced in a `mets:file/mets:Flocat` in the `mets.xml`.
 
+### When in METS and not in data
```
kba: What if those files do not actually exist (like the local `file:/...` links in Goobi/Kitodo output? Or they might not be accessible to checksum?

cneud: METS files that are compatible with the DFG-Viewer ``MUST`` contain a ``<mets:fileGrp USE="DEFAULT">`` that ``MUST`` be resolvable via HTTP so this should be the last resort as a fallback.

cneud:
Closing as this was included with #94.

ocrd_zip.md
kba: The problem is that if we enforce the rule "mets:file in METS -> must be either packaged as payload or listed in fetch.txt" can conflict with the rule "every file listed in fetch.txt must have a checksum and be resolveable". Because references to local files, invalid URLs and non-resolveable URLS are not uncommon in existing METS documents.

cneud: So I assume we can only relax the latter rule "every file listed in fetch.txt must have a checksum and be resolveable"?

kba: No, that rule is part of BagIt https://tools.ietf.org/html/rfc8493#section-2.2.3

cneud:
Need to adapt according to https://github.com/OCR-D/spec/pull/106#discussion_r312170161

kba:
Best to wait for @VolkerHartmann, it's not a high priority issue, concerns only the validators AIFAICS. In the end, it cannot be an error if that rule is broken because real data is messy.

cneud:
Exactly. But would be good if validation does catch this. IME most real data is messy ;)

EEngl52:
@cneud @kba do you still want to adapt this with @VolkerHartmann no longer available?

cneud:
@EEngl52 This has implications on spec (bagit), so we should not close without a solution (even if not currently worked on).

cneud:
100% ACK, but before closing here we should probably still mention this in the relating docs and what about the [validation](https://github.com/OCR-D/spec/pull/106#issuecomment-519629843)?

kba:
You're right, in fact I think we should set `Allow-Fetch.txt: false in the bagit profile and remove any mention of it and partial OCRD-ZIP. BTW the proposed change made it into the spec anyway and has been for two years.

I think the validation is actually easier without fetch.txt because WYSIWYG, no network problems interfering.

bertsky:
Is this related to https://github.com/OCR-D/core/issues/323 by any chance?

kba:
> Is this related to [OCR-D/core#323](https://github.com/OCR-D/core/issues/323) by any chance?

Yes and by extension to https://github.com/OCR-D/core/issues/176

cneud:
Since afaict https://github.com/OCR-D/core/issues/323 is unaffected, this should also be closed with https://github.com/OCR-D/spec/pull/182.

kba:
Superseded by #182

PR #120, opened 2019-07-30
ocrd-tool: add layout/segmentation/text-image

kba:


ocrd_tool.md
```diff
@@ -156,6 +156,7 @@ properties:
                 - recognition/font-identification
                 - recognition/post-correction
                 - layout/segmentation
+                - layout/segmentation/text-image
```
cneud: OK, but some general thoughts:

I have not reviewed the DFKI method yet but in the future we may need sth more abstract than ``text-image`` and/or sth more specific to differentiate between various non-text regions. I am e.g. not certain how current OCR-D segmentation methods segment or classify tables, graphs, formulas? Is a comparable level of segmentation exposed by the other, i.e. Würzburg method?  

What about ``text-other`` ?

kba:
> What about text-other ?

Sounds reasonable. Or even text-nontext? 

> how current OCR-D segmentation methods segment or classify tables, graphs, formulas

DFKI and Würzburg do text/image segmentation, DFKI has block classification in the pipeline. For that maybe `layout/classification`?

Issue #117, opened 2019-07-03
Implement multiple output file groups per processor


wrznr:
Pls. correct me if I am wrong: At the moment, a CLI invocation for an ocrd processor may only specify a single output file group. This is not sufficient for processors which may create multiple output data (i.e. XML and image files) in a single step (e.g. cropping, deskewing). In https://github.com/OCR-D/ocrd_tesserocr/pull/48 this is solved by setting a constant output file group name for the image files. This is a suboptimal solution. Most likely, the `-O` parameter has to be adjusted.

bertsky:
I concur. This would not only be cleaner, but may be the only way to prevent conflicts (if multiple steps in the pipeline want to produce the same _kind_ of image data, but in a different way or from a different source). 

As with multiple input fileGrps, there is a slight complication in that usually, the position in the list might be meaningful and must be documented prominently. For example, one might have to specify `-O OCR-D-SEG-LINE-BIN,OCR-D-IMG-BIN` and _know_ that the second output will be used for the image data, just as in OCR evaluation one might have to specify `-I OCR-D-GT-SEG-LINE,OCR-D-OCR-OCRO,OCR-D-COR-ASV-ANN` and _know_ that the first position will be used as reference/GT.

kba:
> At the moment, a CLI invocation for an ocrd processor may only specify a single output file group

The convention is the same for [--output-file-grp](https://ocr-d.github.io/cli#-o---output-file-grp-grp) as for input file groups, that multiple values can be specified, separated by comma.

Is there something preventing processor implementers from accepting multiple output file groups? Should core offer better support for parsing these comma-separated strings and always provide input/output file groups as an array?

I don't really see how to change the spec to accomodate this.

wrznr:
I guess that if comma-separated values are already available, this issue can be closed and the processors may use them as @bertsky suggested.

kba:
Actually, spec should be adapted to better document this in ocrd-tool.json.

@bertsky:

> Also, what do we do with output_file_grp in ocrd-tool.json? There needs to be a place where assumptions on input and output file groups can be put (how many, what order)... same for multi-OCR input or OCR-GT alignment/evaluation

Tricky because not only can there be multiple output file groups for different data types where order is important, but also different "variants" depending on the input.

Maybe change the structure of `output_file_grp` to a more expressive array of objects, e.g.:

```yaml
"ocrd-tesserocr-binarize": {
  "executable": "ocrd-tesserocr-binarize",
[...]

PR #121, opened 2019-07-31
[WIP] ocrd-tool: change structure of in/output_file_grp param to array of objects

kba:
c.f. #117


ocrd_tool.schema.json
```diff
@@ -1 +1 @@
...
...le: Whether the file should be cached, e.g. because it is large and won't change.","default":false}}}}},"description":{"description":"Concise description what the tool does"},"categories":{"description":"Tools belong to this categories, representing modules within the OCR-D project structure","type":"array","items":{"type":"string","enum":["Image preprocessing","Layout analysis","Text recognition and optimization","Model training","Long-term preservation","Quality assurance"]}},"steps":{"description":"This tool can be used at these steps in the OCR-D functional model","type":"array","items":{"type":"string","enum":["preprocessing/characterization","preprocessing/optimization","preprocessing/optimization/cropping","preprocessing/optimization/deskewing","preprocessing/optimization/despeckling","preprocessing/optimization/dewarping","preprocessing/optimization/binarization","preprocessing/optimization/grayscale_normalization","recognition/text-recognition","recognition/font-identification","recognition/post-correction","layout/segmentation","layout/segmentation/region","layout/segmentation/line","layout/segmentation/word","layout/segmentation/classification","layout/analysis"]}}}}}}}}
```
wrznr: Pretty print would be nice for reviewing!

kba: That file is generated from the [YAML source](https://github.com/OCR-D/spec/pull/121/files/37ea71fa3f13aae5a85be11f44847258dd411463#diff-1e7f24dd18fa27c1cba1e30a749cdf5b) and purposefully minimized to prevent changes. It's documented in the monorepo but not here, some meta-documentation would be helpful I suppose. Sorry for the confusion

kba:
Adding this to the specs would be a breaking change and would require a new major version of spec, core  and all processors. Since we're moving away from separate output file groups for PAGE-XML and AlternativeImages anyway and default values for file groups rarely make sense for general use, the gain in expressiveness is not worth the problems. Closing.

The comments on better documentation of the ocrd-tool.json have merit though, tracking in https://github.com/OCR-D/spec/issues/157.

PR #128, opened 2019-10-31
Fix some typos (found by codespell)

stweil:
Signed-off-by: Stefan Weil <sw@weilnetz.de>

### DIFF
--- a/ocrd_mets.example.xml
+++ b/ocrd_mets.example.xml
@@ -119,7 +119,7 @@
     </mets:digiprovMD>
   </mets:amdSec>
   <mets:fileSec>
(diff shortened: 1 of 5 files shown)


ocrd_mets.example.xml
```diff
@@ -119,7 +119,7 @@
     </mets:digiprovMD>
   </mets:amdSec>
   <mets:fileSec>
-    <!-- The follwing fileGrps are predefined (in the order in which they may appear):
```
kba: This is actually from OCR-D/assets ^^. But I will fix that one quickly myself.

PR #132, opened 2019-12-06
no multi-page TIFF, OCR-D/core#243

kba:


mets.md
```diff
@@ -34,6 +34,14 @@ However, since technical metadata about pixel density is so often lost in
 conversion or inaccurate, processors should assume **300 ppi** for images with
 missing or suspiciously low pixel density metadata.
 
+## No multi-page images
+
+Image formats like TIFF support encoding multiple images in a single file.
+
+Data providers SHOULD provide single-image TIFF.
+
+OCR-D processors MUST ignore any additional images beyond the first.
```
bertsky: This is inconsistent IMO. It invites situations where you loose all but the first pages, yet can claim OCR-D conformance. It shifts the blame to the user.

Why not make single-image a `MUST` for data providers as well? Or what is the context here – at what point does the spec start being effective, before importing data into workspaces or after that?

cneud: In a way, shifting the burden to the user can be considered intentional here, as use of multi-page TIFF for OCR should be discouraged due to many side-effects it can have.

In that sense, I would also be fine with making single-page TIFF a ``MUST`` for input images, which probably requires us to check/warn if that is not the case...

bertsky: > In a way, shifting the burden to the user can be considered intentional here, as use of multi-page TIFF for OCR should be discouraged due to many side-effects it can have.

We seem to have different notions of _user_ here. So for you a user also gets to import the data into OCR-D. (I was under the impression that a _data provider_ would be the person or organisation creating METS/PAGE workspaces.)

> In that sense, I would also be fine with making single-page TIFF a `MUST` for input images, which probably requires us to check/warn if that is not the case...

Exactly: all we have to do _inside_ OCR-D is raise an exception. Now the provider (or user) importing the data gets to fix the problem, but we never risk loosing pages silently.

kba: > to check/warn if that is not the case...

https://github.com/OCR-D/core/pull/369

I'm fine with MUST, multi-page TIFF in a library/archive setting is unusual and probably an error of the data provider.

kba: Changed to "processors MUST raise an exception when encountering multi-page TIFF". I would not want to implement this in core's resolver/workspace though but it will be part of the workspace validation. If the workspace doesn't validate, it should not be processed (OCR-D/core#369)

PR #133, opened 2019-12-11
add metadata directory to allowed tag files

VolkerHartmann:
Bag may contain provenance metadata as tag files. 
Metadata have to be placed in a subdir called 'metadata'.
Subdir may contain any number of '*.xml' and '*.txt' files.


kba:
Fine with me, though `metadata` is not an ideal term IMHO. Can we find a term that conveys that this is about provenance/process metadata? e.g.

* `/prov`
* `/metadata/prov/*`

Issue #134, opened 2019-12-17
Required parameter 'level-of-operation'


kba:
Since this is a setting that concerns most if not all processors, it would IMHO make sense to require processors to implement a parameter `level-of-operation` the way [tesserocr does](https://github.com/OCR-D/ocrd_tesserocr/blob/master/ocrd_tesserocr/ocrd-tool.json#L48-L53).

The possible values would be `page`, `region`, `line`, `word` and `glyph`.

Prrocessors may restrict this further with an enum in the definition for the levels they support and can override the default.

What do you think? @bertsky @wrznr @mjenckel @khurramHashmi

bertsky:
I am much in favour of a uniform parameter name for this in the spec. 

I would even argue that usually processors should provide that parameter (with a default) even in the case where they only allow one specific value/level, so users can quickly assess where they can place them in the workflow – except of course if this is already unambiguous by the name (e.g. `segment-region`) or definition of the operation (e.g. cropping, image characterization)

However, we must be careful with the wording. Currently, we've got:
- `textequiv_level` in
  * `ocrd-tesserocr-recognize` `ocrd-cis-ocropy-recognize` (for the lowest **output** level – input level is then:
     - region, if output level is `region`
     - line, otherwise)
  * `ocrd-keraslm-rate` `ocrd-cor-asv-ann-process` (for **both** input and output level)
    
  This usually affects the levels above that, too!
- `operation_level` in
  * `ocrd-tesserocr-deskew` `ocrd-tesserocr-binarize` `ocrd-anybaseocr-deskew` `ocrd-anybaseocr-dewarp` (with the usual semantics)
  * `ocrd-anybaseocr-block-segmentation` `ocrd-anybaseocr-tiseg` `ocrd-anybaseocr-textline` (still unclear)
- `level-of-operation` in
  * `ocrd-cis-ocropy-binarize` `ocrd-cis-ocropy-denoise` `ocrd-cis-ocropy-deskew` `ocrd-cis-ocropy-clip` `ocrd-kraken-binarize` `ocrd-anybaseocr-binarize` (with the usual semantics)
  * `ocrd-cis-ocropy-segment` (for the **input** level – output level will be 1 lower)

Candidates where that parameter should probably be added, even though they are fixed to 1 level:
- `ocrd-im6convert` (page)
- `ocrd-olena-binarize` (page)
- `ocrd-typegroups-classifier` (page)
- `ocrd-segment-repair` (region)
[...]

PR #136, opened 2020-01-03
define Standard Parameters for dpi, input-level, output-level

kba:
Basis for #134 and OCR-D/core#376

### DIFF
--- a/ocrd_tool.md
+++ b/ocrd_tool.md
@@ -10,6 +10,84 @@ services](swagger).
 
 To validate a `ocrd-tool.json` file, use `ocrd ocrd-tool /path/to/ocrd-tool.json validate`.
 
+## Standard parameters
+
+There is a number of parameters common to all processors that MUST be supported by processors.
+
+### `dpi`
+
+Custom DPI to assume for pixel density of images.
+
+MUST default to 300.
+
+### `input-level`
+
+On what level of typography should input images be processed?
+
+Processors MAY define a `default` value.
+
+`enum` MUST be a list of one or more of:
+
+* `page`
+* `block`
+* `line`
+* `word`
+* `glyph`
+
+### `output-level`
+
+On what level of typography should output images be produced?
+
+Processors MAY define a `default` value.
+
+`enum` MUST be a list of one or more of:
+
+* `page`
+* `block`
+* `line`
+* `word`
+* `glyph`
+
+Whether `input-level` and `output-level` match semantically is up to the
+processor. I.e. if `input-level` and `output-level` are inconsistent according
+to its semantics, processors MUST refuse further processing.
+
+### Sample for standard parameters
+
+Here is a snippet of an `ocrd-tool.json` for a tool that can operate on `page`, `block` or `line` level
+and produce output on `block`, `line` or `glyph` level, e.g. [ocrd-cis-ocropy-segment](https://github.com/cisocrgroup/ocrd_cis/blob/dev/ocrd_cis/ocropy/segment.py):
+
+```hjson
+{
+  [...]
+  "parameter": {
+    "dpi": {
+      "type": "number",
+      "default": 300,
+    },
+    "input-level": {
+      "type": "string":
+      "enum": ["page", "block", "line"],
+      "default": "page"
+    }
+    "output-level": {
+      "type": "array":
+      "item": {
+        "type": "string",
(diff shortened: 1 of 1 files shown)


ocrd_tool.md
```diff
@@ -10,6 +10,84 @@ services](swagger).
 
 To validate a `ocrd-tool.json` file, use `ocrd ocrd-tool /path/to/ocrd-tool.json validate`.
 
+## Standard parameters
+
+There is a number of parameters common to all processors that MUST be supported by processors.
```
bertsky: I thought the idea was actually to provide extra parameters __automatically__ to all processors (without any need for specification/validation and implementation in the individual processors). So this would have to enter the spec at CLI.md, not here – right?

And besides using extra parameters, I have [proposed](https://github.com/OCR-D/core/issues/376#issuecomment-562349028) other mechanisms (like extra CLI options, and environment variables), too. Shouldn't we at least discuss these options first?

ocrd_tool.md
```diff
@@ -10,6 +10,84 @@ services](swagger).
...
+
+### `dpi`
+
+Custom DPI to assume for pixel density of images.
```
bertsky: I have already [proposed](https://github.com/OCR-D/core/issues/376#issuecomment-567390606) a better solution for DPI overrides: by static annotation in the respective PAGE-XML attributes. `OcrdExif` would then have to prefer that over the binary meta-data if available. And dedicated processors could estimate DPI from the image, or allow user overrides. I thought you were going along with that?

ocrd_tool.md
```diff
@@ -10,6 +10,84 @@ services](swagger).
...
+
+MUST default to 300.
+
```
bertsky: A _default_ here? If you don't make this optional, then trusting the binary image meta-data density value becomes impossible.

ocrd_tool.md
```diff
@@ -10,6 +10,84 @@ services](swagger).
...
+Processors MAY define a `default` value.
+
+`enum` MUST be a list of one or more of:
+
+* `page`
+* `block`
+* `line`
+* `word`
+* `glyph`
+
```
bertsky: IMHO we should only have `output-level`. See [arguments here](https://github.com/OCR-D/spec/issues/134#issuecomment-570662568).

ocrd_tool.md
kba: I do. I'd still offer it as a standard parameter with `parameter > annotation > EXIF metadata`.

ocrd_tool.md
kba: Hmm, the point here was to have a place for that discussion. I wrote down what came to mind at the place where it fitted best. I wouldn't want to necessarily force everyone to implement a `dpi` parameter they do not need (core might help here though). Could be overrideable via env var, getopt-style extension etc. We just need a solution soon and a PR is more urgent than issues.

> So this would have to enter the spec at CLI.md, not here – right?

Again, I'm open for changing the details, as long as there is one place with a fragid `#standard-parameters` that describes these "special" "parameters"/"flags"/"settings".

ocrd_tool.md
kba: Good point. But we need a fallback *somewhere* to activate if
* no param provided
* no annotation in PAGE
* none or garbage EXIF data

ocrd_tool.md
bertsky: Ah, I see. Makes sense (as long as we make sure these standard parameters are really automatic for tools).

ocrd_tool.md
bertsky: Indeed. The only place I have seen in the spec for that was your recent addition to mets.md:

> However, since technical metadata about pixel density is so often lost in
conversion or inaccurate, processors should assume **300 ppi** for images with
missing or suspiciously low pixel density metadata.

In practise, this is all just a contract around behaviour in core's OcrdExif (or whatever that will be named after we respect the PAGE-XML overrides).

ocrd_tool.md
kba: I removed the section for input-level here and added an example what behavior should be like if the expected input-level isn't provided.

It's probably more productive to continue that discussion on input/output-level in https://github.com/OCR-D/spec/issues/134. I've assgned it to the reviewers here, so let's settle that first and then I'll update the PR.

PR #141, opened 2020-01-31
ocrd-tool: allow defining "sample-calls" to illustrate usage of tools

kba:
We're all rather frantically trying out permutations of different tools with different parameters etc. With this change, tool developers can provide sample calls of their tool, e.g. with different parameters, multiple input file groups etc. We can reasonably require of developers that these calls when made on e.g. the workspaces in assets should succeed, so we could automate these as smoke tests in e.g. ocrd_all to avoid regressions.

Plus, it's really helpful documentation.


kba:
I would also be okay with a more elaborate syntax to show how tools interact with others, e.g. in the form of task sequences as used by `ocrd process`.

EEngl52:
From a user's perspective I would find it very helpful to have sample calls for each processor - especially when they indicate parameters (as e.g. a model) which always have to be set. It would certainly lower the threshold for beginners if you could just copy and paste those calls or at least see what they should look like.

wrznr:
Both reviewers express valid points: Sample calls are helpful but only if you have the means to run them. Since this is a PR to the specification, I think they should be added nonetheless if only to motivate the processor maintainers to make their invocations and models more transparent.

kba:
I added a commit that doesn't show up [because GitHub's infrastructure issues today](https://www.githubstatus.com/incidents/q07bfjh7jf1t):

```diff
 diff --git a/ocrd_tool.md b/ocrd_tool.md
 index adb2f80..add1da5 100644
 --- a/ocrd_tool.md
 +++ b/ocrd_tool.md
 @@ -172,11 +172,25 @@ properties:
                    - call
                    - description
                  properties:
 +                  description:
 +                    description: What this command call does
 +                    type: string
                    call:
 +                    description: The command line call
                      type: string
                      pattern: '^ocrd-'
 -                  description:
 +                  ocrd_zip:
 +                    description: OCRD-ZIP fitting the command line call
 +                    type: string
 +                    format: uri
 +                  input-image:
 +                    description: A visual representation of an input image/document before processing
 +                    type: string
 +                    format: uri
 +                  output-image:
 +                    description: A visual representation of an output image/document after processing
                      type: string
 +                    format: uri
```

This would allow defining an OCRD-ZIP as the workspace to call the command on (e.g. from our GT repo or OCR-D/assets), as well as before/after image representations to illustrate to users what impact the processor has.

> For many tools, this would require placing a "standard" model right into the repo itself (not just a submodule or URL or allowing training).

As the focus can at first be more on the documentation, an easy solution could be to add this in a comment within the call:

```yaml
sample-calls:
  - description: Executes the foo processor with the bar model
    call: '''
      # Download the model first:
      # wget 'https://model-repo/bar'
      ocrd-foo -p '{"model": "bar"}'
    ocrd-zip: https://gt-repo/sample.ocrd.zip
    input-image: https://image-host/before.png
    output-image: https://image-host/after.png
```

cneud:
Sorry @bertsky I accidentally clicked on "request another round of review" from you and it seems there is no way to undo it. Plz ignore!

cneud:
While I appreciate the concerns expressed [here](https://github.com/OCR-D/spec/pull/141#pullrequestreview-351785858), I assume the (1) point is mitigated somwhat by the coming-soon ``make models`` target whereas the (2) point is partially addressed by [this comment](https://github.com/OCR-D/spec/pull/141#issuecomment-592040144). 

So imho there is sufficient benefit for end users from this to add it as a convention/recommendation.

CHANGELOG.md
```diff
@@ -139,6 +143,7 @@ Added:
 
   * docker: instructions on naming and labelling images, #139
   * CLI tools must implement `-h/--help`, #115
+>>>>>>> master
```
mweidling, line 146:
```suggestion

```

mweidling (commented):
This is an unnoticed merge relic, I guess.

Issue #147, opened 2020-04-17
require processors to be non-destructive


bertsky:
IINM there is no place in the spec yet which explicitly states that processors must not overwrite or delete existing annotations.

I was under the impression that this is a core principle of the functional model. If so, it should be documented (with binding language).

For example, processors should not delete their output files if they already exist.

Strictly, that also goes for secondary data like derived images. (But sadly, we have `force=True` by default for `workspace.save_image_file` in core already.)

This place in the spec could also pave the way for exceptions (under control of the user) like the [`--overwrite` addition](https://github.com/OCR-D/core/issues/434) to the CLI.

kba:
Fair point. Where shall we add this info? CLI would make sense but doesn't really fit or would be hard to find. METS would also fit because there's already invariants on processor behavior in there. Or we could create a distinct page on processors.

bertsky:
> Where shall we add this info? CLI would make sense but doesn't really fit or would be hard to find.

I would say: doesn't really fit _yet_. See #148 

> METS would also fit because there's already invariants on processor behavior in there.

IMO the METS spec does not really fit because it's the static/data perspective.

PR #148, opened 2020-04-20
explicate missing long-standing conventions...

bertsky:
- introduce term _processor_ properly
- clarify mandatory/optional w.r.t. conforming
- add cross-references to METS and ocrd-tool specs
- add statement about strictly functional / non-destructive interface
- clarify that no `-g` means all pages
- clarify that no `-p` means default/no parameters


bertsky:
Try to fix #147

bertsky:
I wonder why there is no reference to logging.md yet ...

Issue #158, opened 2020-06-09
More powerful parameter handling


kba:
Currenly, parameters can be passed to processors as one JSON string or one JSON filename. We should make that more flexible:

1. Make the `parameter` option repeatable and merge the JSON documents from right to left. This allows simple inheritance where more generic parameter files, possibly bundled with a processor, can be overridden with more specific values.

2. Allow more user-friendly declaration of paramers on the command-line. cf. https://github.com/OCR-D/core/issues/345

bertsky:
Speaking of – how about:

 3. Allow comments within parameter JSON files for (sharing/storing) self-documenting parameter files.
   Example:
    ```python
    {
     # set non-zero to override DPI setting in meta-data;
     # can influence automatic window_size below
     "dpi": 0,
     # window tiles always need to include _some_ fg text
     # otherwise they will magnify background noise;
     # because input images can have different DPI though,
     # it's hard to configure a good fixed window size.
     # setting zero here will make the processor use
     # a DPI-dependent (odd-valued) window size
     "window_size": 0,
     # lower values increase the fg weight; this can
     # help preserve small/thin text features and thus
     # retain glyphs as mostly connected components,
     # but may require binary denoising afterwards to
     # remove small fg specks
     "k": 0.1
    }
    ```
    To parse this in Python, one can simply do:
    ```python
    import json,re
    with open('param.json', 'r') as file:
        text = file.read()
        text = re.sub('\s#.*', '', text, flags=re.MULTILINE)
        result = json.loads(text)
    ```

bertsky:
If parameter files can be combined, then perhaps we could also distribute examples of them with the module packages! But how does one access them from the CLI? 

What if we allowed referencing installed parameter files by file name (like a relative path), and when looking up local files we include the module directory (besides the CWD) as search path?

Example:

    ocrd-preprocess-image -I OCR-D-IMG -O OCR-D-SEG-PAGE-DEN-NRM,OCR-D-IMG-DEN-NRM -p param_im6convert-denoise-raw.json

…where `param_im6convert-denoise-raw.json` has been installed along with `ocrd_wrap` into its site-packages directory.

Then we should also list the available parameter files in the help text, e.g.:
```
Usage: ocrd-preprocess-image [OPTIONS]

  Convert or enhance images

Options:
  -V, --version                   Show version
  -l, --log-level [OFF|ERROR|WARN|INFO|DEBUG|TRACE]
                                  Log level
[...]

Issue #160, opened 2020-06-10
search paths for file references in processor parameters


bertsky:
Since #69 there is the attributes `content-type` and `cacheable` in the parameter json, dedicated to file resources. These could be config files or model files the processor needs.

Now naturally, a problem arises if these file paths in the parameter json are encapsulated into parameter or workflow files (to be distributed/shared across installations): you obviously cannot use absolute path names for portability, but there's no specification on where the processor should search for relative path names.

There are many scenarios to be found among current implementations:
1. no provisions at all: Python will only search the CWD, which is always the workspace directory. This will fail, unless the user or workflow engine copies or symlinks the resource files there.
2. add to or replace CWD with the parent directory of `__file__`: this is not always predictable, because it depends on how the package was installed, but can work when the model files are installed into the package directory.
3. look up some specific environment variable (`TESSDATA_PREFIX`, `OCROPUS_DATA`, `CORASVANN_DATA`)
4. look up pre-defined fixed paths

IMO We need some common convention here.

2 is complicated by the fact that you usually cannot distribute model files as part of `package_data` via PyPI or even in the same repo, because they are too big. But the processor should be able to expect some _installer_ (presumably within ocrd_all) to download and copy them _somewhere_ for it to find at runtime under a relative path.

3 should be considered a stop-gap solution.

4 could also be something like `sysconfig.get_path('data')/share/ocrd_models` (which should point to `$VIRTUAL_ENV`)

This may even be useful for distributing/sharing (perhaps partial) [parameter files](https://github.com/OCR-D/spec/issues/158#issuecomment-641782548).

bertsky:
This has been fixed in specification (CLI: resource handling, file-type parameters for model resources and parameter files for presets; ocrd-tool.json: file-type parameter relative pathname resolution) and implementation. Happy ending!

PR #162, opened 2020-06-14
CLI: Introduce parameter presets, #158

kba:
The intention here is to simplify re-using parameter settings. Processors can bundle parameter setting presets, e.g. using a specific model or a specific combination of values. Users can store such presets at certain locations in the file system. A preset name can be used instead of a JSON string or filename for the `-p`/`--parameter` CLI option.

This is a first draft, bundled presets should probably described in `ocrd-tool.json` and lookup path simplified, etc.

### DIFF
--- a/cli.md
+++ b/cli.md
@@ -134,13 +134,13 @@ other implementation-specific means of logging configuration. For example, with
 Instead of processing METS, output the [ocrd-tool](ocrd_tool) description for
 this executable, in particular its parameters.
 
-### `-C, --preset PRESET_NAME`
+### `-C, --show-asset FILENAME`
 
-Print the contents of preset `PRESET_NAME`. Look up the resp. JSON file according to [file parameter lookup rules](ocrd_tool#file-parameter)
+Print the contents of processor asset `FILENAME`. Look up the resp. absolute filename according to the [file parameter lookup rules](ocrd_tool#file-parameter).
 
-### `-L, --presets`
+### `-L, --list-assets`
 
-List the names of [presets](#presets).
+List the names of [processor-assets](#processor-assets) in all of the paths defined by.the [file parameter lookup rules](ocrd_tool#file-parameter), one name per line.
 
 ### `-h, --help`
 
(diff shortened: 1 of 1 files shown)


kba:
> bundled with a processor

means: stored as `package_data` and manageable with `pkg_resources`.

> But couldn't we handle --presets (listing) via normal --help?

Yes, that would be the human-readable version. `--presets` would only list the names of presets one-per-line for automation. I have a branch on the backburner that adds shell completion, that would be an application.

`sysconfig.get_path('data') will be equal to `$VIRTUAL_ENV` when using a venv/virtualenv (and we require it).

>  Perhaps we should first find a general convention for #160 and then just refer to that for preset files.

Fair point. I'll add a section on parameter filename resolution and we can agree on a search path first and then complete this PR.

kba:
The file-parameter-resolution PR is now merged, I think this one is ready to ship as well (and then implement it in core for processors to make use of asap).

@bertsky Objections?

bertsky:
> The file-parameter-resolution PR is now merged, I think this one is ready to ship as well (and then implement it in core for processors to make use of asap).
> 
> @bertsky Objections?

No, looks good!

Just that maybe we don't even need an extra `-C|--preset` now: If we generally require resolving relative path names for parameter files to adhere to the new ocrd-tool fpath conventions, then we can make the modules install their parameter files under these locations and users can just point to them via regular `-p|--parameter`. We would not even have to change any logic, only the documentation (including the term _preset_ itself). That would only leave `-L|--presets` as new. (But then again, why not also have a command to list all installed models under these paths?)

kba:
We can use the same mechanism for presets and general file parameters, that is true (and should be reflected by referencing the resolution logic in ocrd_tool.md). *Preset* is limiting. "Bundled files" perhaps? "Included batteries" - something like that? Maybe just "assets"?

A generalized `-L/--list-bundled` (or whatever the wording) is still useful for automation, as is `-C/--dump-asset` (again: open on the wording). Having a mechanism short of actually running a processor to get to the files it bundles (or has available in the paths defined in the "file parameters" section in ocrd_tool), is still a sensible idea, again, for automation, generating user interfaces, things like dynamically loading a model into a memory-mapped FS location before processing etc.

kba:
I've changed the text to be more generic. I chose `processor asset` as the working title.

bertsky:
> We can use the same mechanism for presets and general file parameters, that is true

Sorry, I got confused. Of course we can – that's we have been planning. The `-C|--preset` was merely the dump action.

> (and should be reflected by referencing the resolution logic in ocrd_tool.md)

Perhaps the current text under the `Presets` section can be reformulated in a way a preset is nothing more than a relative filename for a bundled parameter file.

> _Preset_ is limiting. "Bundled files" perhaps? "Included batteries" - something like that? Maybe just "assets"?
> A generalized `-L/--list-bundled` (or whatever the wording) is still useful for automation, as is `-C/--dump-asset` (again: open on the wording). Having a mechanism short of actually running a processor to get to the files it bundles (or has available in the paths defined in the "file parameters" section in ocrd_tool), is still a sensible idea, again, for automation, generating user interfaces, things like dynamically loading a model into a memory-mapped FS location before processing etc.

Yes, absolutely. (I did not intend to argue against dump and list commands, see above.)

Okay, I get that it makes sense to subsume presets and models in both commands. One can still differentiate them by name / filename suffix I guess. 

For the wording, I would vote for _bundled files_, _resource files_, or _package resources_. _Assets_ is too much financial jargon to my taste (although I know it's become common via App programming). For example, `-L|--list-resources` and `-C|--dump-resource`.

cli.md
```diff
@@ -168,14 +168,22 @@ The log messages must have the format `TIME LEVEL LOGGERNAME - MESSAGE\n`, where
...
+filename by following the [conventions laid out in the `ocrd_tool`
+spec](ocrd_tool#file-parameters). These files, either bundled by the processor
+developer or put in place by the user, are called *processor assets*. The
+*processor assets* of a processor can be listed with the `-L/--list-assets`
+option and individual *processor assets* can be retrieved with the
+`-C/--show-asset` option. Since *processor assets* use the same mechanism
+as file parameters, they can be used
+
+  * as the argument to the `-p/--parameter` option, and
+  * as the value of a file parameter
```
bertsky, line 183:
```suggestion
  * as the argument to the `-p/--parameter` option (i.e. a **preset** file), and
  * as the value of a file-type parameter (e.g. a **model** file)
```

kba:
> Assets is too much financial jargon to my taste (although I know it's become common via App programming)

The joy of a second language, I was not aware of either when creating `OCR-D/assets`...

I like `processor resources`.

PR #163, opened 2020-06-15
ocrd_tool: file parameter relative filename resolution order, #160

kba:


ocrd_tool.md
```diff
@@ -19,6 +19,27 @@ Optionally, workflow processors can be notified that this file is potentially
...
 indefinitely after download by setting the `cacheable` property to `true`.
 
+The filename itself, i.e. the concrete value `<fpath>` of a file parameter
+should be resolved in the following way:
+
+* If `<fpath>` is an `http`/`https` URL: Download to a temporary directory (if
+  `cacheable==False`) or a semi-temporary cache directory (if `cacheable==True`)
+* If `<fpath>` is an absolute path: Use as-is.
+* If `<fpath>` is a relative path, try resolving the following paths and return
+  the first one found if any, raise a `FileNotFoundError` otherwise:
```
bertsky, line 29: I wouldn't be too Python-specific here. Maybe `abort with an error` suffices...

ocrd_tool.md
```diff
@@ -19,6 +19,27 @@ Optionally, workflow processors can be notified that this file is potentially
...
+  * `$CWD/<fpath>`
```
bertsky, line 30: I would welcome this from a user perspective, but it might be harder to implement (because we loose CWD context as soon as we `chdir` to the workspace after processor invocation).

ocrd_tool.md
```diff
@@ -19,6 +19,27 @@ Optionally, workflow processors can be notified that this file is potentially
...
+  * If an environment variable is defined that has the name of the processor in
+    upper-case and with `-` replaced with `-` and followed by `_PATH` (e.g. for a processor
+    `ocrd-dummy`, the variable would need to be called `OCRD_DUMMY_PATH`):
+    * Split the variable value at `:` and try to resolve by prepending each token
+      with `<fname>`
```
bertsky, line 35: I don't understand this. What is `fname` here?

ocrd_tool.md
```diff
@@ -19,6 +19,27 @@ Optionally, workflow processors can be notified that this file is potentially
...
+  * `$VIRTUAL_ENV/share/<name-of-processor>/<fpath>`
+  * `$HOME/.local/share/<name-of-processor>/<fpath>`
+  * `$HOME/.config/<name-of-processor>/<fpath>`
```
bertsky, line 38: Since this looks like [XDG Base Directory standard](https://unix.stackexchange.com/a/313001), maybe allow overriding these by `$XDG_DATA_HOME` and `$XDG_CONFIG_HOME`?

ocrd_tool.md
```diff
@@ -19,6 +19,27 @@ Optionally, workflow processors can be notified that this file is potentially
...
+* If `<fpath>` is an absolute path: Use as-is.
+* If `<fpath>` is a relative path, try resolving the following paths and return
+  the first one found if any, raise a `FileNotFoundError` otherwise:
+  * `$CWD/<fpath>`
+  * If an environment variable is defined that has the name of the processor in
+    upper-case and with `-` replaced with `-` and followed by `_PATH` (e.g. for a processor
+    `ocrd-dummy`, the variable would need to be called `OCRD_DUMMY_PATH`):
+    * Split the variable value at `:` and try to resolve by prepending each token
+      with `<fname>`
+  * `$VIRTUAL_ENV/share/<name-of-processor>/<fpath>`
```
bertsky, line 36: This one is quite Python-specific, and even venv-specific (which our Docker images don't use). Plus this would already be (functionally) equivalent to the last list item, wouldn't it?

ocrd_tool.md
```diff
@@ -19,6 +19,27 @@ Optionally, workflow processors can be notified that this file is potentially
...
+  * `$HOME/.local/share/<name-of-processor>/<fpath>`
+  * `$HOME/.config/<name-of-processor>/<fpath>`
+  * `/usr/local/share/<name-of-processor>/<fpath>`
+  * If it is a Python implementation: Replace `/` with `.` in `<fpath>` and try to
```
bertsky, line 40:
```suggestion
  * If it is a Python implementation: Replace leading `/` with `.` in `<fpath>` and try to
```

ocrd_tool.md
```diff
@@ -19,6 +19,27 @@ Optionally, workflow processors can be notified that this file is potentially
...
+    resolve wirth `pkg_resources.resource_filename`
```
bertsky, line 41: How about `sysconfig.get_path('data') + '/share'`?

ocrd_tool.md
kba: True, though I've never seen anyone change either, ever and

`~/.config` and even
`$HOME/.config`
is shorter than
`$XDG_CONFIG_HOME`.

But you're right, I'm a big fan of XDG, should call them as such.

ocrd_tool.md
kba: It is Python-specific but local ocrd_all will install also the non-python code to `$VIRTUAL_ENV`. The really python-specific variant would be `os.path.join(sysconfig.get_path('data'), 'share', '<name-of-processor>', '<fpath>))`. And python-specific is okay IMHO because it offers python developers an additional way to bundle data with a processor while still allowing users to store their processor-specific files in clearly defined locations.

ocrd_tool.md
kba: No, I actually mean `s,/,.,g`. But that's nonsense, `pkg_resources.resource_filename(__name__, "<fpath>")` will handle path separators correctly.

A leading `/` would mean it should be interpreted as an absolute file.

ocrd_tool.md
kba: Was from a previous revision, should read:

> Split the variable value at : and try to resolve by appending `<fpath>`
> to each token and return the first found file if any.

ocrd_tool.md
kba: That is indeed a bit tricky in OCR-D/core but that should not prevent us. It's a common sense convention and we will find a way to implement it in core.

ocrd_tool.md
kba: That is not processor-specific. Do we need a processor-independent location to look for files?

We explicitly want to encourage processor developers to bundle smaller files like parameter presets, mapping tables etc. directly with their code, so it is distributed on PyPI, handled by standard Python tooling. It's non-trivial to determine where these files are actually stored w/o Python and in such a setting, it's better to store them at one of the previous locations.

ocrd_tool.md
bertsky: Oh, now I got it. List-valued like `PATH`

ocrd_tool.md
bertsky: But I did not mean _replace_ with the variable expansion, only _try_ it first. (IIUC, XDG requires to allow these as override for the fixed path name. They don't need to have any value. They don't have on Ubuntu.)

ocrd_tool.md
bertsky: I fully agree.

ocrd_tool.md
kba: we're in agreement. if the wording is unclear, I'm open to suggestions.

ocrd_tool.md
bertsky: Sorry, I didn't mean it that way. I was still with `$VIRTUAL_ENV/share/<name-of-processor>` above, only accessing it by its canonical name (via `sysconfig`) not `VIRTUAL_ENV`.

But using `resource_filename` it better (more Pythonic) then some `share` convention anyway.

ocrd_tool.md
```diff
@@ -26,19 +26,18 @@ should be resolved in the following way:
...
     `ocrd-dummy`, the variable would need to be called `OCRD_DUMMY_PATH`):
-    * Split the variable value at `:` and try to resolve by prepending each token
-      with `<fname>`
+    * Split the variable value at `:` and try to resolve by appending `<fpath>`
+      to each token and return the first found file if any
   * `$VIRTUAL_ENV/share/<name-of-processor>/<fpath>`
-  * `$HOME/.local/share/<name-of-processor>/<fpath>`
-  * `$HOME/.config/<name-of-processor>/<fpath>`
+  * `$XDG_DATA_HOME/<name-of-processor>/<fpath>` (by default, `$XDG_DATA_HOME == $HOME/.local/share`)
+  * `$XDG_CONFIG_HOME/<name-of-processor>/<fpath>` (by default, `$XDG_CONFIG_HOME == $HOME/.config`)
```
bertsky, line 38: less ambiguous wording:
```suggestion
  * `$XDG_DATA_HOME/<name-of-processor>/<fpath>` (with `$HOME/.local/share` instead of `$XDG_DATA_HOME` if unset)
  * `$XDG_CONFIG_HOME/<name-of-processor>/<fpath>` (with `$HOME/.config` instead of `$XDG_CONFIG_HOME` if unset)
```

ocrd_tool.md
kba: thx

PR #164, opened 2020-06-16
Alternative image same folder

kba:
https://github.com/OCR-D/core/issues/505

The convention "all files in a filegroup must have the same mimetype" is obsolete.

AlternativeImages should be written to the same file group as the PAGE-XML they are referenced in first.

### DIFF
--- a/mets.md
+++ b/mets.md
@@ -91,7 +91,7 @@ with the type of manipulation (`BIN-KRAKEN`).
 `<mets:fileGrp USE="OCR-D-SEG-WORD">`       | Word segmentation
 `<mets:fileGrp USE="OCR-D-SEG-GLYPH">`      | Glyph segmentation
 `<mets:fileGrp USE="OCR-D-OCR-TESS">`       | [Tesseract OCR](https://github.com/OCR-D/ocrd_tesserocr)
-`<mets:fileGrp USE="OCR-D-OCR-ANY">`        | [AnyOCR](https://github.com/OCR-D/ocrd_anybaseocr)
+`<mets:fileGrp USE="OCR-D-OCR-OCRO">`        | [Ocropus OCR](https://github.com/OCR-D/ocrd_cis)
 `<mets:fileGrp USE="OCR-D-COR-CIS">`        | [CIS post-correction](https://github.com/cisocrgroup/ocrd_cis)
 `<mets:fileGrp USE="OCR-D-COR-ASV">`        | [ASV post-correction](https://github.com/ASVLeipzig/cor-asv-ann)
 `<mets:fileGrp USE="OCR-D-GT-SEG-REGION">`   | Region segmentation ground truth
--- a/mets.md
+++ b/mets.md
@@ -54,10 +54,6 @@ For this purpose, the METS file MUST contain a `mods:identifier` that must conta
 * `url`
 
 
-## File Group 
-
-All `mets:file` inside a `mets:fileGrp` MUST have the same `MIMETYPE`.
-
 ## File Group USE syntax
 
 All `mets:fileGrp` MUST have a **unique** `USE` attribute that hints at the provenance of the files.
--- a/mets.md
+++ b/mets.md
@@ -106,10 +106,7 @@ The `ID` MUST be unique inside the METS file.
 
 ```
 FILEID := ID + "_" + [0-9]{4}
-ID := "OCR-D-" + WORKFLOW_STEP + ("-" + FEATURE)* + ("-" + PROCESSOR)?
-FEATURE := ("BIN" | "CROP" | "DESKEW" | DESPECK" | "DEWARP" )
-WORKFLOW_STEP := ("IMG" | "SEG" | "OCR" | "COR")
-PROCESSOR := [A-Z0-9\-]{3,}
+ID := FILEGRP + (".IMG")?
 ```
 ### Examples
 
--- a/mets.md
+++ b/mets.md
@@ -106,7 +106,8 @@ The `ID` MUST be unique inside the METS file.
 
 ```
 FILEID := ID + "_" + [0-9]{4}
-ID := "OCR-D-" + WORKFLOW_STEP + ("-" + PROCESSOR)?
+ID := "OCR-D-" + WORKFLOW_STEP + ("-" + FEATURE)* + ("-" + PROCESSOR)?
+FEATURE := ("BIN" | "CROP" | "DESKEW" | DESPECK" | "DEWARP" )
 WORKFLOW_STEP := ("IMG" | "SEG" | "OCR" | "COR")
 PROCESSOR := [A-Z0-9\-]{3,}
 ```
@@ -116,6 +117,7 @@ PROCESSOR := [A-Z0-9\-]{3,}
 --               | --
 `<mets:file ID="OCR-D-IMG_0001">`            | The unmanipulated source image
 `<mets:file ID="OCR-D-IMG-BIN_0001">`        | Black-and-White image
+`<mets:file ID="OCR-D-IMG-BIN-CROP_0001">`   | Cropped Black-and-White image
 
 ## Grouping files by page
 
--- a/mets.md
+++ b/mets.md
@@ -86,24 +86,14 @@ with the type of manipulation (`BIN-KRAKEN`).
 `<mets:fileGrp USE>` | Type of use for OCR-D
 --                                          | --
 `<mets:fileGrp USE="OCR-D-IMG">`            | The unmanipulated source images
-`<mets:fileGrp USE="OCR-D-IMG-BIN">`        | Black-and-White images
-`<mets:fileGrp USE="OCR-D-IMG-CROP">`       | Cropped images
-`<mets:fileGrp USE="OCR-D-IMG-DESKEW">`     | Deskewed images
-`<mets:fileGrp USE="OCR-D-IMG-DESPECK">`    | Despeckled images
-`<mets:fileGrp USE="OCR-D-IMG-DEWARP">`     | Dewarped images
 `<mets:fileGrp USE="OCR-D-SEG-REGION">`      | Region segmentation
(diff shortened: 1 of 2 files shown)


page.md
```diff
@@ -24,7 +24,11 @@ The `imageFilename` of the `<pg:Page>` and `filename` of the
 `<pg:AlternativeImage>` element MUST be a filename relative to the `mets.xml`.
 
 All URL used in `imageFilename` and `filename` [MUST be referenced in a fileGrp
-in METS](https://ocr-d.de/en/spec/mets#if-in-page-then-in-mets).
+in METS](https://ocr-d.de/en/spec/mets#if-in-page-then-in-mets). This MUST be
+the same file group as the PAGE-XML that was the result of the processing step
+that produced the `<pg:AlternativeImage>`. In other words:
+`<pg:AlternativeImage>` should be written to the same fileGrp as its source
+PAGE-XML, which in most implementations will mean the same folder.
```
bertsky, line 31: There is a mirroring formulation in the METS.md [here](https://ocr-d.de/en/spec/mets#if-in-page-then-in-mets) which should be modified likewise.

kba: I've included your proposals, thx.

mets.md
```diff
@@ -86,24 +86,14 @@ with the type of manipulation (`BIN-KRAKEN`).
...
-`<mets:fileGrp USE="OCR-D-OCR-ANY">`        | AnyOCR
-`<mets:fileGrp USE="OCR-D-COR-CIS">`        | CIS post-correction
-`<mets:fileGrp USE="OCR-D-COR-ASV">`        | ASV post-correction
-`<mets:fileGrp USE="OCR-D-GT-IMG-BIN">`     | Black-and-White images ground truth
-`<mets:fileGrp USE="OCR-D-GT-IMG-CROP">`    | Cropped images ground truth
-`<mets:fileGrp USE="OCR-D-GT-IMG-DESKEW">`  | Deskewed images ground truth
-`<mets:fileGrp USE="OCR-D-GT-IMG-DESPECK">` | Despeckled images ground truth
-`<mets:fileGrp USE="OCR-D-GT-IMG-DEWARP">`  | Dewarped images ground truth
+`<mets:fileGrp USE="OCR-D-OCR-TESS">`       | [Tesseract OCR](https://github.com/OCR-D/ocrd_tesserocr)
+`<mets:fileGrp USE="OCR-D-OCR-ANY">`        | [AnyOCR](https://github.com/OCR-D/ocrd_anybaseocr)
```
bertsky, line 94:
```suggestion
`<mets:fileGrp USE="OCR-D-OCR-OCRO">`        | [Ocropus OCR](https://github.com/OCR-D/ocrd_cis)
```

mets.md
```diff
@@ -86,24 +86,14 @@ with the type of manipulation (`BIN-KRAKEN`).
 `<mets:fileGrp USE>` | Type of use for OCR-D
 --                                          | --
 `<mets:fileGrp USE="OCR-D-IMG">`            | The unmanipulated source images
-`<mets:fileGrp USE="OCR-D-IMG-BIN">`        | Black-and-White images
-`<mets:fileGrp USE="OCR-D-IMG-CROP">`       | Cropped images
-`<mets:fileGrp USE="OCR-D-IMG-DESKEW">`     | Deskewed images
-`<mets:fileGrp USE="OCR-D-IMG-DESPECK">`    | Despeckled images
-`<mets:fileGrp USE="OCR-D-IMG-DEWARP">`     | Dewarped images
```
bertsky, line 97: These conventions still make sense for the file ID naming scheme within fileGrps, so they should **re-appear** there.

(See comment in [my first review](https://github.com/OCR-D/spec/pull/164#pullrequestreview-431636277).)

bertsky: Sorry, didn't see the other commits!

bertsky:
I am not so sure anymore about the new `mets:file ID` scheme – even if it's merely a _SHOULD_. If we recommend a fixed ID scheme for processors, then we will again end up with a much increased likelihood of conflicting IDs when workflows are more complex/elaborate. IMHO the best recommendation for _processors_ is to always base their new IDs on the existing ones dynamically, merely replacing substrings (input fileGrp by output fileGrp) or adding suffixes. So choosing unique names would ultimately be up to the import-time ID choices, and processors would never have to be creative/imaginative. The longer and more complex the workflow, the longer the IDs.

kba:
>  IMHO the best recommendation for processors is to always base their new IDs on the existing ones dynamically, merely replacing substrings (input fileGrp by output fileGrp) or adding suffixes.

Agree in principle and as I said in https://github.com/cisocrgroup/ocrd_cis/pull/57, would even recommend we offer quality-of-life functions for that in ocrd_utils.

However, I see a problem with the initial data. E.g. @EEngl52 ran into an issue today where the file names and `mets:file/@ID` are purely numeric. To circumvent the invalid `mets:file/@ID` (which mustn't begin with a number), the fix was to add some prefix (e.g. `IMG_`), not necessarily matching the input file group. And in this case, we can control the `ocrd workspace (bulk-)add` calls but what if we start with METS that doesn't follow the "input-file-group-as-part-of-file-id" convention? We need a fallback for that - or am I missing something?

bertsky:
> alls but what if we start with METS that doesn't follow the "input-file-group-as-part-of-file-id" convention? We need a fallback for that - or am I missing something?

I should have been more precise: yes, for cases where the ID does not contain the fileGrp, one should fall back to some form of `concat_padded(output_file_grp, n)`, as done in https://github.com/cisocrgroup/ocrd_cis/pull/57. This is still unique and doesn't require the processor to be imaginative.

kba:
I think the wording on file IDs - SHOULD rather than MUST - should give enough leeway for producing sensible file IDs. With https://github.com/OCR-D/spec/pull/162 we introduce  `make_file_id` in core which should further simplify the process.

So I think this is ready for merge as-is.

bertsky:
> I think the wording on file IDs - SHOULD rather than MUST - should give enough leeway for producing sensible file IDs. With #162 we introduce `make_file_id` in core which should further simplify the process.

But `make_file_id` does not help you for derived images (which this discussion is all about). You can use it to get a new ID for the output PAGE from the input image/PAGE, but not for an output image directly. 

So do we recommend:
- just adding a suffix to that (e.g. `.IMG-BIN`), which is fully dynamic and never clashes, or
- using a fixed ID scheme for derived images in the processor implementation, which is static and therefore might clash again.

IMHO even if it's just a recommendation, not a requirement, we should describe the very practice that we are also going implement (for good reasons), not some artificial "aestheticised" scheme.

mets.md
```diff
@@ -120,7 +106,8 @@ The `ID` MUST be unique inside the METS file.
 
 ```
 FILEID := ID + "_" + [0-9]{4}
-ID := "OCR-D-" + WORKFLOW_STEP + ("-" + PROCESSOR)?
+ID := "OCR-D-" + WORKFLOW_STEP + ("-" + FEATURE)* + ("-" + PROCESSOR)?
+FEATURE := ("BIN" | "CROP" | "DESKEW" | DESPECK" | "DEWARP" )
 WORKFLOW_STEP := ("IMG" | "SEG" | "OCR" | "COR")
 PROCESSOR := [A-Z0-9\-]{3,}
```
bertsky, line 112:
```suggestion
ID := FILEGRP + (".IMG")?
```

mets.md
bertsky: How about introducing `PRE` as a workflow step? This is different than `IMG` (which is merely the original input image) and different than the old recommendation for separate image file groups. It would be used for PAGE and derived images alike (discerning the latter only with an ID suffix, see below).

```suggestion
`<mets:fileGrp USE="OCR-D-PRE-BIN">`        | Binarization preprocessing
`<mets:fileGrp USE="OCR-D-PRE-CROP">`       | Cropping preprocessing
`<mets:fileGrp USE="OCR-D-PRE-DESKEW">`     | Deskewing preprocessing
`<mets:fileGrp USE="OCR-D-PRE-DESPECK">`    | Despeckling preprocessing
`<mets:fileGrp USE="OCR-D-PRE-DEWARP">`     | Dewarping preprocessing
```

mets.md
```diff
@@ -130,6 +117,7 @@ PROCESSOR := [A-Z0-9\-]{3,}
 --               | --
 `<mets:file ID="OCR-D-IMG_0001">`            | The unmanipulated source image
 `<mets:file ID="OCR-D-IMG-BIN_0001">`        | Black-and-White image
+`<mets:file ID="OCR-D-IMG-BIN-CROP_0001">`   | Cropped Black-and-White image
```
bertsky, line 120:
```suggestion
`<mets:file ID="OCR-D-IMG_0001">`            | The unmanipulated source image
`<mets:file ID="OCR-D-PRE-BIN_0001">`        | PAGE encapsulating the result from binarization
`<mets:file ID="OCR-D-PRE-BIN.IMG_0001">`    | Black-and-white image
`<mets:file ID="OCR-D-PRE-CROP_0001">`       | PAGE encapsulating the result from (binarization and) cropping
`<mets:file ID="OCR-D-PRE-CROP.IMG_0001">`   | Cropped black-and-white image
```

mets.md
kba: > artificial "aestheticised" scheme

😆 I just love pseudo-backus-naur grammars, but I see your point.

`ID := FILEGRP + "-IMG"` makes sense.

bertsky: > > artificial "aestheticised" scheme
> 
> :laughing: I just love pseudo-backus-naur grammars, but I see your point.

I am with you: it's a shame letting pretty BNFs that look natural go to waste. But users and implementors have been avoiding that scheme. We now have the fortunate opportunity to introduce something simpler that is also more workable.

> `ID := FILEGRP + "-IMG"` makes sense.

Glad you agree. Do you want me to replace `.` with `-` in my suggestions, or are you going to reformulate anyway?

mets.md
kba: @tboenig @EEngl52 what do you think?

PR #169, opened 2020-07-28
glossary: define "processor"

kba:


kba:
> We should be clear wrt to the distinction of "software"/"tools"/MP and "methods"/"activities" exposed by individual `Processors`. If my wording of "method" contradicts with other places where we call these "activities" that is also fine, although I think "method" is most specific.

I know the semantics of the concepts here a bit too well perhaps, if "method" makes it clearer IYHO, sure.

PR #175, opened 2020-10-12
allow array values for parameters

kba:
Since we have full JSON parsing in place for `-P` and `-p`, there is no reason not to support array parameter values.


bertsky (approved):
Splendid, thanks!

PR #177, opened 2021-01-07
revise processor resource spec, OCR-D/core#559

kba:
~~I also haven't implemented the `/usr/local/share/<name-of-processor>` variant in OCR-D/core#559 yet - do we still need it?~~

* Drop the pkg_resources stuff
* Drop `/usr/local/share`
* Add intermediary directory `ocrd-resources` to make it easier to mount a host-folder of resources into a Docker image e.g.


bertsky (approved):

bertsky:
> I also haven't implemented the `/usr/local/share/<name-of-processor>` variant in [OCR-D/core#559](https://github.com/OCR-D/core/pull/559) yet - do we still need it?

It would now be the only place with an absolute path, I'd say scratch that too.

PR #181, opened 2021-03-17
add resource schema and resource_locations to ocrd-tool.json schema

kba:
Adds two new properties to the `tools` list items in `ocrd-tool.json`:

* `resources`: List resources that can be used with this processor, to be used with OCR-D/core's `ocrd resmgr` CLI and resource lookup mechanism
* `resource_locations`: By default, resource's are looked up in `data` (`$XDG_DATA_HOME`), `cwd` and `system` (`/usr/local/share/ocrd-resources`). Since not all processors can support the full range of locations (notably ocrd_tesserocr can look only in one location for models), this allows restricting the locations to look for this processor.

Until now, we're maintaining a [central resource registry](https://github.com/OCR-D/core/blob/master/ocrd/ocrd/resource_list.yml) in OCR-D/core. We want to decentralize this, allowing processor developers to list resources in their own ocrd-tool.json.

### DIFF
--- a/ocrd_tool.schema.yml
+++ b/ocrd_tool.schema.yml
@@ -126,3 +126,52 @@ properties:
                 - layout/segmentation/word
                 - layout/segmentation/classification
                 - layout/analysis
+          resource_locations:
+            type: array
+            description: The locations in the filesystem this processor supports for resource lookup
+            default: ['data', 'cwd', 'system']
+            items:
+              type: string
+              enum: ['data', 'cwd', 'system']
+          resources:
+            type: array
+            description: Resources for this processor
+            items:
+              type: object
+              additionalProperties: false
+              required:
+                - url
+                - description
+                - name
+                - size
+              properties:
+                url:
+                  type: string
+                  description: URLs of all components of this resource
+                description:
+                  type: string
+                  description: A description of the resource
+                name:
+                  type: string
+                  description: Name to store the resource as
+                type:
+                  type: string
+                  enum: ['file', 'github-dir', 'tarball']
+                  default: file
+                  description: Type of the URL
+                parameter_usage:
+                  type: string
+                  description: Defines how the parameter is to be used
+                  enum: ['as-is', 'without-extension']
+                  default: 'as-is'
+                path_in_archive:
+                  type: string
+                  description: if type is archive, the resource is at this location in the archive
+                  default: '.'
+                version_range:
+                  type: string
+                  description: Range of supported versions, syntax like in PEP 440
+                  default: '>= 0.0.1'
+                size:
+                  type: number
+                  description: Size of the resource in bytes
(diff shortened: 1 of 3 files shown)


bertsky:
Should probably be addressed here, too: 
- resources of type `directory` (see [discussion about possible representation in parameters](https://github.com/OCR-D/core/issues/750#issuecomment-999728361))
- resources that reside in the distribution directory of the module and come pre-installed (think Python site-packages).

ocrd_tool.schema.yml
```diff
@@ -126,3 +126,52 @@ properties:
...
+                  description: URLs of all components of this resource
+                description:
+                  type: string
+                  description: A description of the resource
+                name:
+                  type: string
+                  description: Name to store the resource as
+                type:
+                  type: string
+                  enum: ['file', 'github-dir', 'tarball']
```
bertsky, line 159:
```suggestion
                  enum: ['file', 'directory', 'archive']
```

Should be renamed in existing implementation (`resource_manager.py`, `resource_list.yml`), too.

bertsky:
> * resources that reside in the distribution directory of the module and come pre-installed (think Python site-packages).

there should be no need to _register_ these, though

ocrd_tool.schema.yml
```diff
@@ -126,3 +126,52 @@ properties:
                 - layout/segmentation/word
                 - layout/segmentation/classification
                 - layout/analysis
+          resource_locations:
+            type: array
+            description: The locations in the filesystem this processor supports for resource lookup
+            default: ['data', 'cwd', 'system']
+            items:
+              type: string
+              enum: ['data', 'cwd', 'system']
```
bertsky, line 135: Even if we would not need to **register** these concretely under `resources` in the tool json, the processor should specify under `resource_locations` whether it allows for (or provides) any pre-installed resources – **module** location:
```suggestion
          resource_locations:
            type: array
            description: The locations in the filesystem this processor supports for resource lookup
            default: ['data', 'cwd', 'system', 'module']
            items:
              type: string
              enum: ['data', 'cwd', 'system', 'module']
```

bertsky (commented):
Also, the ocrd_tool.md needs to be updated as well:
- [ ] mention the `resources` and `resource_locations` under the `File parameters` section
- [ ] show them in the examples

ocrd_tool.schema.yml
kba: `github-dir` is only a stub in `resource_manager.py`, not implemented. Maybe we can decide on how to move forward with directory resources to replace `github-dir` with `directory`. Then we probably should also distinguish between files and directories within an archive. So far, we only have directories in archives (calamari, eynollah etc) but this might change in the future.

bertsky: > Then we probably should also distinguish between files and directories within an archive. So far, we only have directories in archives (calamari, eynollah etc) but this might change in the future.

We have `path_in_archive` though. So even if you only target a file in an archive, it should already be covered.

PR #182, opened 2021-05-31
ocrd_zip: drop Manifestation-Depth, disallow fetch.txt

kba:
This PR removes the `Ocrd-Manifestation-Depth` parameter and disallows the `fetch.txt` mechanism.

We introduced these to allow for iterative ingestions of only the changed files into OCR-D GT endpoints like the [OCR-D GT Repo](http://ocr-d.de/gt-repo) or [OLA-HD](https://www.sub.uni-goettingen.de/projekte-forschung/projektdetails/projekt/ola-hd-ein-ocr-d-langzeitarchiv-fuer-historische-drucke/). 

However, I think this flexibility is a premature optimization. Yes, workspaces can become very large and ingesting full manifestations for every update is inefficient. But bandwidth has not been an issue so far and it will be difficult to map these mechanisms to the (messy) real-life data we want to process, e.g. with hard-to-categorize `@xlink:href` (think `file:/` URL from Goobi/Kitodo).

I therefore think it would be better if we focussed on packaging all the OCR-D produced data in a well-defined way and ensure that data consumers don't have to do any extra steps (that might fail!) to create a complete manifestation. "What you see is what you get" is more important than maximum efficiency for re-ingestion.

We should still have such a mechanism for updating an ingested OCRD-ZIP but we should find a more efficient and less ambiguous way POSTing an an incomplete bag (such as a set of patches against the contents of the OCRD-ZIP or an API to PUT specific results in the OCRD-ZIP).


cneud:
> this flexibility is a premature optimization

I wholeheartedly agree. Let's keep it simple but safe.

If there are no objections from the perspective of [OLA-HD](https://www.sub.uni-goettingen.de/projekte-forschung/projektdetails/projekt/ola-hd-ein-ocr-d-langzeitarchiv-fuer-historische-drucke/) this has my full support.

cneud:
> this flexibility is a premature optimization

I wholeheartedly agree. Let's keep it simple but safe.

If there are no objections from the perspective of [OLA-HD](https://www.sub.uni-goettingen.de/projekte-forschung/projektdetails/projekt/ola-hd-ein-ocr-d-langzeitarchiv-fuer-historische-drucke/) this has my full support.

PR #183, opened 2021-09-02
logging should only go to STDERR

kba:
Having output to be parsed on STDOUT and logging on STDERR consistenly will ensure that logging never interferes with parseable output, e.g. from `ocrd-name-of-processor --dump-json` or in bashlib-based calls.

### DIFF
--- a/cli.md
+++ b/cli.md
@@ -159,11 +159,9 @@ Successful execution should signal `0`. Any non-zero return value is considered 
 
 ## Logging
 
-Data printed to `STDERR` and `STDOUT` is captured linewise and stored as log data.
+Data printed to `STDERR` is captured linewise and stored as log data.
 
 Processors must adjust logging verbosity according to the [`--log-level` parameter](#-l---log-level-loglevel).
-
-Errors, especially those leading to exceptions, must be printed to `STDERR`.
 
 The log messages must have the format `TIME LEVEL LOGGERNAME - MESSAGE\n`, where
 


bertsky:
I'm in on that, but I'm also afraid this is still no guarantee against interference on stdout: processors might call legacy code which simply cannot be silenced or made to adhere to our logging conventions. (We could perhaps redirect _all_ stdout to stderr, but then the question becomes: what _is_ allowed on stdout actually?)

kba:
> what _is_ allowed on stdout actually?

I would say anything that is to be parsed by software, such as the `--dump-json` ocrd-tool.json or "one item per line" CSV output of `ocrd workspace find`.

bertsky:
> I would say anything that is to be parsed by software, such as the `--dump-json` ocrd-tool.json or "one item per line" CSV output of `ocrd workspace find`.

Agreed! So we could really redirect stdout to stderr, but prior to that, duplicate stdout to some dedicated FD, which we then use for printing/dumping the things which the command is actually made for, right?

So we still need to go through all output of the `ocrd` (sub)commands and the processor decorator, and decide which actually belongs to stdout...

kba:
> So we still need to go through all output of the `ocrd` (sub)commands and the processor decorator, and decide which actually belongs to stdout...

Agreed, but AFAICS this should not prevent merging this and OCR-D/core#713 in the meantime, right?

bertsky:
> Agreed, but AFAICS this should not prevent merging this and [OCR-D/core#713](https://github.com/OCR-D/core/pull/713) in the meantime, right?

Not at all – just trying to figure out what this would mean for our implementation.

(I do find it adequate for a spec like this to underspecify what should or should not actually appear on stdout. It's not needed for interoperability, so why be too restrictive. But in our implementation we can try to be rigorous, and thus fix its potential stdout parsing problem once and for all.)

PR #184, opened 2021-11-01
Resource lookup: for `--location cwd` look directly in $CWD, no subdirectory, OCR-D/core#727

kba:


ocrd_tool.md
```diff
@@ -27,7 +27,7 @@ should be resolved in the following way:
 * If `<fpath>` is an absolute path: Use as-is.
 * If `<fpath>` is a relative path, try resolving the following paths and return
   the first one found if any, otherwise abort with an error message stating so:
-  * `$CWD/ocrd-resources/<fpath>`
+  * `$CWD/<fpath>` (**Note** that the file is expected to be directly be in `$CWD`, not in a subdirectory)
```
bertsky, line 30: Shouldn't we use `$PWD` here, the conventional/POSIX/bash/ksh/zsh envvar, instead of `CWD`? Otherwise, better define as meta-variable and then write as `<cwd>/<fpath>` or similar...

kba: I'm happy with `$PWD`, changed in 0c2922e

However we still have `--location cwd` for `ocrd resmgr`, should we change as well? I am not super-keen to do that since reference `cwd`/`CWD` in quite a few places...

bertsky: I was merely concerned that users which understand shell variable syntax (and thus know `$PWD` but not `$CWD`) would get confused. But you are right: users that don't recognise this would be better off with a reference in terms of `<cwd>`, after introducing it properly in the text.

kba: ae8cab4

bertsky: Splendid!

bertsky (approved):

PR #187, opened 2021-11-24
remove front matter from glossary.de.md, atx header for changelog

### LINKED ISSUE #186: add markdown preamble with title to all the markdown files
Since we're publishing the spec with Jekyll, we should add a [front matter](https://jekyllrb.com/docs/front-matter/) with the page title to all the markdown files here and adapt our [publishing setup](https://github.com/OCR-D/ocrd-website/blob/master/Makefile#L170-L174) accordingly.

kba:
Makes the markdown files consistent to fix the publishing in https://github.com/OCR-D/ocrd-website. 

@lena-hinrichsen The `glossary.de.md` is not actually translated. I apparently forgot to check in the actual changes when adding this file. Maybe we should just drop it until we get around to translating? In any case, it is now possible to translate files of the spec to english by copying the file and replacing `.md` with `.de.md`. These will be placed in the right spot for publishing in ocrd-website's `make spec`.

fix #186


lena-hinrichsen:
sounds good!

PR #189, opened 2021-12-22
ocrd-tool: require content-type=text/directory for directory parameters

kba:
See https://github.com/OCR-D/core/issues/750 for discussion.

Tangentially related: https://github.com/OCR-D/core/issues/691

### DIFF
--- a/ocrd_tool.schema.yml
+++ b/ocrd_tool.schema.yml
@@ -126,3 +126,15 @@ properties:
                 - layout/segmentation/word
                 - layout/segmentation/classification
                 - layout/analysis
+          resource-content-type:
+            type: string
+            default: application/octet-stream
+            enum: ['text/directory', 'application/octet-stream']
+            description: >
+              The format of resources this processor expects. Most processors
+              use files for resources (e.g. `*.traineddata` for
+              `ocrd-tesserocr-recognize`) while others use directories of files
+              (e.g. `default` for `ocrd-eynollah-segment`).  If a processor
+              requires directories, it must set `resource-content-type` to
+              `text/directory`.
+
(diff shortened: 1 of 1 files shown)


bertsky:
Note: what I meant by…
> But then, why not just use `content-type: text/directory` or `format: directory`?

…was to use the **existing** `parameters/properties` content-type or format.

ocrd_tool.schema.yml
```diff
@@ -83,11 +83,12 @@ properties:
...
-                      The format of resources this processor expects for this
-                      parameter. Most processors use files for resources (e.g.
-                      `*.traineddata` for `ocrd-tesserocr-recognize`) while
-                      others use directories of files (e.g. `default` for
+                      The media type of resources this processor expects for
+                      this parameter. Most processors use files for resources
+                      (e.g.  `*.traineddata` for `ocrd-tesserocr-recognize`)
+                      while others use directories of files (e.g. `default` for
                       `ocrd-eynollah-segment`).  If a processor requires
                       directories, it must set `resource-content-type` to
```
bertsky, line 93:
```suggestion
                      directories, it must set `content-type` to
```

ocrd_tool.schema.yml
```diff
@@ -83,7 +83,15 @@ properties:
...
                   content-type:
                     type: string
-                    description: "If parameter is reference to file: Media type of the file"
+                    default: 'application/octet-stream'
+                    description: >
+                      The media type of resources this processor expects for
+                      this parameter. Most processors use files for resources
+                      (e.g.  `*.traineddata` for `ocrd-tesserocr-recognize`)
+                      while others use directories of files (e.g. `default` for
+                      `ocrd-eynollah-segment`).  If a processor requires
```
bertsky, line 92:
```suggestion
                      `ocrd-eynollah-segment`).  If a parameter requires
```

PR #190, opened 2022-02-04
ocrd_tool: add module directory

bertsky:
As implemented in https://github.com/OCR-D/core/pull/792

(Also contains a clarification about subdirs as used in Tesseract e.g. `script/Fraktur`.)


kba:
merged into #181

PR #199, opened 2022-04-26
minor fixes for web-api

cneud:
* typo corrected ("iff" -> "if")
* use https for OCR-D website and Apache license

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -9,7 +9,7 @@ info:
 
     ## OCR-D API compatibility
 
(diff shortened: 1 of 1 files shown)


mweidling:
I think the `iff` was intended by @kba because it's a logical term meaning "if and only if". But it confused others as well, so we might write it more explicitly (@kba)?

cneud:
Yeah I was unsure about that too, but since this is in documentation and not actual code...let's see what @kba thinks.

kba:
Yeah, it was meant as "if and only if" - I like precise language :) But since it confuses you, let's drop it.

PR #203, opened 2022-04-27
fix list item visibility in ocrd_zip docu

joschrew:
without this addition the whole bullet point for Ocrd-Mets is not visible in the resulting html


cneud (approved):
LGTM!

PR #205, opened 2022-05-09
added missing change for bagit-profile-json-url

joschrew:
I think one occurrence of bagit-profile url was overlooked during the last change regarding the same topic.


kba:
Indeed! Well spotted, thanks.

PR #206, opened 2022-05-20
ocrd-tool.json: allow more fine-grained parameter descriptions

kba:
cf. OCR-D/core#848

Allows processor developers to describe numerical, object and array parameters in more detail.

Drops the pattern on input/output file groups since the naming convention is a `SHOULD` not a `MUST`.


bertsky (approved):
Will these be validated automatically by any chance?

PR #209, opened 2022-07-06
webapi-spec: content of put/post for workspaces

joschrew:
In my opinion the content/body for post and put a workspace have to be the same. But I'm not 100 % sure how to define them. I read this https://swagger.io/docs/specification/describing-request-body/file-upload/ and that: https://stackoverflow.com/questions/14455408/how-to-post-files-in-swagger-openapi.
I want to use `WorkspaceRequest` defined in component.schema, but I think the json-part should be left out. So this PR is what I came up with.

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -165,7 +165,7 @@ paths:
           description: Return Log
           content:
             'text/plain': {}
-        '404': 
+        '404':
           description: 'ProcessorJob not found'
 
   '/workflow':
@@ -282,7 +282,8 @@ paths:
       requestBody:
         description: OCRD-ZIP of the new workspace
         content:
-          'application/vnd.ocrd+zip': {}
+          'application/vnd.ocrd+zip':
+            schema: {$ref: '#/components/schemas/WorkspaceRequest'}
         required: true
       responses:
         '201':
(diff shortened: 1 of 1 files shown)


openapi.yml
```diff
@@ -282,7 +282,8 @@ paths:
       requestBody:
         description: OCRD-ZIP of the new workspace
         content:
-          'application/vnd.ocrd+zip': {}
+          'application/vnd.ocrd+zip':
+            schema: {$ref: '#/components/schemas/WorkspaceRequest'}
```
kba, line 286: If you really don't want the upload to be a multi-part form (descriptive JSON + binary workspace), then I suggest to get rid of the `WorkspaceRequest` altogether and instead make this a simple file upload.

```suggestion
          'application/vnd.ocrd+zip':
            schema:
              type: string
              format: binary
```

BTW, you can upload the OpenAPI YAML definition to https://app.swaggerhub.com and edit interactively to see how the constructs would be rendered by the Swagger UI. I have adapted https://app.swaggerhub.com/apis/kba/ocr-d_web_api/0.0.1#/workspace/createWorkspace to show how this would be rendered. If you click on the "Try it Out" button, you should see a file upload form element.

kba (changes requested):

joschrew:
Ok, thank you. I now updatet the PR to what I think it should look like regarding the post end put request. As you said I (really) would prefer a simple file upload for both methods without additional json. Thanks for pointing out how to view the api online, it helped me.

PR #220, opened 2022-08-15
openapi: Add requestBody to runWorkflow

joschrew:
Currently, I think, we only need the workspace_id (and the workflow_id provided in the path) to run a workflow. I think this is always required for other implementers as well. Please correct me if I am wrong. But I think that it is likely, that we or other implementers may encounter another parameter needed, like the starting file-grp for example. Because of that, I decided to add the `parameters` section like in processorArgs. I think this gives more flexibility to the implementers. Because I am not too sure about that please don't hesitate to comment if you have another opinion on that.

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -454,7 +454,10 @@ components:
       required: ['workspace_id']
       properties:
         workspace_id: {type: string}
-        parameters:
+        workflow_parameters:
+          description: >-
+            Possibility to add additional parameters for the workflow. The parameters that can/must
+            be specified here depend on the respective implementation of the workflow.
           type: object
           default: {}
     WorkflowJob:
--- a/openapi.yml
+++ b/openapi.yml
@@ -231,6 +231,11 @@ paths:
           description: ID of the Workflow
           schema: {type: string}
           required: true
+      requestBody:
+        description: Execute this Workflow
+        content:
+          application/json: {schema: {$ref: '#/components/schemas/WorkflowArgs'}}
+        required: true
       responses:
         '200':
           description: Return WorkflowJob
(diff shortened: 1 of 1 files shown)


openapi.yml
```diff
@@ -231,6 +231,11 @@ paths:
           description: ID of the Workflow
           schema: {type: string}
           required: true
+      requestBody:
```
kba, line 234: True, we need a way to pass (OCR-D) parameters along with the workspace-id and the requestBody is the best place (in theory this could also be sent via URI encoded URL query params but that is ugly).

openapi.yml
```diff
@@ -443,6 +448,15 @@ components:
     Workflow:
       allOf:
         - {$ref: '#/components/schemas/Resource'}
+    WorkflowArgs:
+      description: The arguments needed to run the Workflow
+      type: object
+      required: ['workspace_id']
+      properties:
+        workspace_id: {type: string}
```
kba, line 456: This seems redundant, since the `workflow_id` is already in the request URL. We should either remove it from the endpoint (`POST /workspace/{workspace-id}` -> `POST /workspace`) or from the `WorkflowArgs, otherwise there's a risk of inconsistency.

kba (changes requested):

openapi.yml
joschrew: I think you are confusing "workflow" and "workspace", please correct me if I am wrong. The `workflow_id` is submitted via the url-path and the `workspace_id` is submitted in the json-body.
But maybe you are still right and it would be better to remove `workspace_id` and expect it in the `properties`-Object. But as I thought the workspace_id is required always, I set it as a separate parameter.

joschrew:
I am unsure about the description but couldn't find a better one

kba (approved):
LGTM, the description is clear enough too IMHO.

PR #221, opened 2022-08-16
CLI: add // regex prefix for page-id

bertsky:
Implemented in core long ago, and useful.


kba (approved):
Indeed, this should be documented, thx.

PR #222, opened 2022-08-19
Web api

tdoan2010:
This is my first attempt for the Web API documentation. Comments are welcome.


lena-hinrichsen (approved):

web_api.md
```diff
@@ -0,0 +1,186 @@
...
+# alternatively using Docker
+docker run --rm -u $(id -u) -v $PWD:/data -w /data -- ocrd/all:maximum ocrd-[processor needed] -I [Input-Group] -O [Output-Group] -P [parameter]
+```
+
+Although this approach works perfectly fine, it poses a drawback: the processor cannot be called remotely. One always
+needs to connect to the machine where the processor is installed and execute the command on the CLI. If the data is
+small, this might not be a problem. However, in the era of big data, where one has to constantly deal with a huge amount
+of input, resource utilization has become vital and a distributed system is the way to go. To achieve a distributed
+OCR-D system, the mentioned drawback must be overcome. Therefore, we provide a wrapper for processors to enable their
+communication over network ability.
```
bertsky, line 152:
```suggestion
Obviously, command-line interfaces do not suit all use-cases well: Calling processors remotely would require 
connecting to the machine via some login and data transfer mechanism, and then executing the CLI locally. Also,
the initialization cannot be easily shared between runs, and that overhead might be very large for GPU processing.
Moreover, scaling up can be difficult: A machine may only support a certain number of concurrent computations,
so it is usually necessary to control the maximum number of parallel sessions.

With a server interface, all these shortcomings can be overcome: Remote access, preloading, queueing etc. become possible.
Also, in virtualized execution environments, containers can only call each other as network services. 
So deploying single-processor containers as network services allows building larger systems.
```

web_api.md
```diff
@@ -0,0 +1,186 @@
...
+
+```shell
+# 1. Via processor CLI
+ocrd-dummy --server=0.0.0.0:80:mongodb://localhost:27017
+
+# 2. Via ocrd CLI
+ocrd server --server=0.0.0.0:80:mongodb://localhost:27017 ocrd-dummy
+```
+
+**Note**: this feature is currently under code review. Once it is finished and release, this page will be updated with
```
bertsky, line 185:
```suggestion
**Note**: this feature is currently under code review. Once it is finished and released, this page will be updated with
```

I am confused: Where do you actually describe the REST interface? Shouldn't we see the description of `GET /` and `POST /` and `GET /{jobid}` here?

bertsky (commented):
Mostly formulations, but some issues too.

web_api.md
```diff
@@ -0,0 +1,261 @@
...
+   or `path_to_privkey` (path to a private key). All Processing Servers, which will be deployed, must be declared under
+   the `deploy_processors` property. In case `type` is `docker`, make sure that [Docker](https://www.docker.com/) is
+   installed in the target machine and the provided `username` has enough rights to execute Docker commands.
+
+Among three sections, only the `message_queue` is required. However, if `hosts` is presented, `mongo_db` must be there
+as well. For more information, please check the [configuration file schema](web_api/config.schema.yml).
+
+### Processing Server
+
+One do not need to start (or stop) a Processing Server manually, since it can be managed by a Processing Broker via
```
kba, line 176:
```suggestion
There is normally no need to start (or stop) a Processing Server manually, since it can be managed by a Processing Broker via
```

web_api.md
```diff
@@ -0,0 +1,261 @@
...
+a [configuration file](#processing-broker). However, if it is necessary to do so, there are two ways to start a
+Processing Server:
+
+```shell
+# 1. Use ocrd tool
```
kba, line 181:
```suggestion
# 1. Use ocrd CLI bundled with OCR-D/core
```

web_api.md
```diff
@@ -0,0 +1,261 @@
...
+
+### Message Queue
+
+By using a queuing system, specifically [RabbitMQ](https://www.rabbitmq.com/), the reliability and flexibility of the
+whole system are greatly improved. In our implementation, manual acknowledgement mode is used. This means, when a
+Processing Server finishes successfully, it sends a positive ACK signal to RabbitMQ. In case of failure, it tries again
+three times before sending a negative ACK signal. When a negative signal is received, RabbitMQ will re-queue the
+message. If there is not any ACK signal sent for any reason (e.g. consumer crash, power outage, network problem, etc.),
+RabbitMQ will automatically re-queue the message after timeout, which is 30
+minutes [by default](https://www.rabbitmq.com/consumers.html#acknowledgement-timeout). This value is, however, can be
```
kba, line 200:
```suggestion
minutes [by default](https://www.rabbitmq.com/consumers.html#acknowledgement-timeout). This behavior can be [overridden](link that describes how].
```

web_api.md
```diff
@@ -0,0 +1,261 @@
...
+configured.
```
kba, line 201:
```suggestion
```

web_api.md
```diff
@@ -0,0 +1,261 @@
...
+
+To avoid processing the same input twice (in case of re-queuing), a Processing Server first checks
+the [`redeliver`](https://www.rabbitmq.com/confirms.html#automatic-requeueing) property to see if this message was
+re-queued. If yes, and the status of this process in the database is not `SUCCESS`, it will process the data described
+in the message again.
+
+When a Processing Broker receives a request, it adds some more data to the content, then push it to an appropriate
```
kba, line 208: >it adds some more data to the content

What is "some more data" here?

kba: Or rather, as it is explained later, maybe remove this here to reduce confusion.

web_api.md
```diff
@@ -0,0 +1,261 @@
...
+```
+
+In the message content, except `job_id`, `processor_name`, and `created_time`, are added by the Processing Broker. The
+rest comes from the body of the `POST /processor/{executable}` request.
+
+Instead of `path_to_mets`, one can also use `workspace_id` to specify a workspace. An ID of a workspace can be obtained
+from the Workspace Server. In case `result_queue_name` property is presented, the result of the processing will be
+pushed to the queue with the provided name. If the queue does not exist yet, it will be created on the fly. This is
+useful when there is another service waiting for the results of processing. That service can simply listen to that queue
+and will be immediately notified when the results are available. The message schema can be
```
kba, line 238:
```suggestion
and will be immediately notified when the results are available. The message schema for results can be
```

web_api.md
```diff
@@ -0,0 +1,261 @@
...
+With the returned `job_id`, one can retrieve more information by sending a `GET` request to
+the `/processor/{executable}/{job_id}` endpoint, or to `/processor/{executable}/{job_id}/log` to get all logs of that
+job.
+
+### Database
+
+The database is required to store necessary information such as users requests, jobs statuses, workspaces, etc. We
+recommend to use [MongoDB](https://www.mongodb.com/) since it is used by Processing Servers, but other kinds of storage
+may work as well. However, there must be a MongoDB running in the system because all Processing Servers will read and
+write to it. To connect to MongoDB via a Graphical User
```
kba, line 256: This seems contradictory. If MongoDB is required for the broker, why and how could you use "other kinds of storage"?

kba (changes requested):
LGTM in general, makes it much clearer how the components work together for me.

web_api.md
```diff
@@ -0,0 +1,261 @@
...
+    * Trigger a single processor on a workspace by calling the `POST /processor/{executable}` endpoint with the chosen
+      processor name and workspace ID, or
+    * Start a workflow on a workspace by calling the `POST /workflow/{workflow-id}` endpoint with the chosen workflow ID
+      and workspace ID.
+    * In both case, a job ID is returned.
+5. With the given job ID, it is possible to check the job status by calling:
+    * `GET /processor/{executable}/{job-id}` for a single processor, or
+    * `GET /workflow/{workflow-id}/{job-id}` for the workflow.
+6. The result can be downloaded by calling the `GET /workspace/{workspace-id}` endpoint with the
+   header `Accept: application/vnd.ocrd+zip`. Without that header, only the metadata of the specified workspace is
```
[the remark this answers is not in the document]
tdoan2010: OK, I agree.

web_api.md
```diff
@@ -0,0 +1,261 @@
...
+
+### Description
+
+As shown in Fig. 1, each section in the [Web API specification](#the-specification) is implemented by different servers,
+which are Discovery Server, Processing Broker, Workflow Server, and Workspace Server respectively. Although each server
+in the figure is deployed on its own machine, it is completely up to the implementors to decide which machines run which
+servers. However, having each processor run on its own machine reduces the risk of version and resource conflicts.
+Furthermore, the machine can be customized to best fit the processor's hardware requirements and throughput demand. For
+example, some processors need GPU computation, while others do not, or some need more CPU capacity while others need
+more memory. It is also easier to scale up the processors, or even apply Function-as-a-Service on some of them, which
```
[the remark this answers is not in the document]
tdoan2010: I remove it then.

web_api.md
```diff
@@ -0,0 +1,261 @@
...
+are not constantly used, to save resources.
+
+**Processing**: since the `Processing` section is provided by [OCR-D Core](https://github.com/OCR-D/core), implementors
+do not need to implement Processing Broker, Message Queue, and Processing Server. Once a request comes, the broker
+pre-processes it if necessary, and push it to an appropriate queue. A processing queue always has the same name as its
```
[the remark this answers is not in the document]
tdoan2010: As explained later, it just converts the request body into a message and push it to the queue.

web_api.md
```diff
@@ -0,0 +1,261 @@
...
+respective processors. For example, `ocrd-olena-binarize` processors listen only to the queue
+named `ocrd-olena-binarize`. A Processing Server, which is
+an [OCR-D Processor](https://ocr-d.de/en/spec/glossary#ocr-d-processor) running as a worker, listens to the queue, pulls
+new jobs when available, processes them, and returns results. One do not call a Processing Server directly, but via a
```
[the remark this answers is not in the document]
tdoan2010: It pushes the status of the job back to the queue. I will re-phrase it to make it clearer.

web_api.md
```diff
@@ -0,0 +1,261 @@
...
+Processing Broker. Job statuses can be pushed back to the queue, depending on the [job configuration](#message-queue),
+so that other services get updates and act accordingly.
+
+**Database**: in this architecture, a database is required to store necessary information such as users requests, jobs
+statuses, workspaces, etc. We recommend to use [MongoDB](https://www.mongodb.com/) since it is used by Processing
+Servers, but other kinds of storage may work as well.
```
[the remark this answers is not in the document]
tdoan2010: Yes, you're right.

web_api.md
```diff
@@ -0,0 +1,261 @@
...
+
+**Network File System**: in order to avoid file transfer between different machines, it is highly recommended to have
+a [Network File System](https://en.wikipedia.org/wiki/Network_File_System) set up. With NFS, all Processing Servers
+(specifically processors) can work in a shared storage environment and access files as if they are local files. To get
+data into the NFS, one could use the `POST /workspace` endpoint to upload [OCRD-ZIP](https://ocr-d.de/en/spec/ocrd_zip)
+files. However, this approach is only appropriate for small data size. Usually, Workspace Server should be able to pull
+data from other storages.
```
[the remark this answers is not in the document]
tdoan2010: The detail implementation is up to the IP. In my mind, the endpoint accepts a URL (and credentials if necessary), then just download the zip file and extract it to create a workspace.

IPs can also implement that endpoint in a way that it accepts a URL to a METS file, download it, then download all files in METS to create a workspace.

I don't include this endpoint in the `openapi.yml` file, since I don't think it should be compulsory. IPs can decide if they want to implement it or not.

web_api.md
```diff
@@ -0,0 +1,261 @@
...
+
+### Processing Broker
+
+A Processing Broker is a server which exposes REST endpoints in the `Processing` section of
+the [Web API specification](openapi.yml). There are two types of task performed by a broker: deployment management and
+message producer. For the former, a broker can deploy, re-use, and shutdown Processing Servers, Message Queue, and
```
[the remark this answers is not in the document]
tdoan2010: Because they're all components in the architecture. I didn't think much about it, actually.

web_api.md
```diff
@@ -0,0 +1,261 @@
...
+      - name: ocrd-eynollah-segment
+        number_of_instance: 1
+        deploy_type: native
+```
+
+There are three main sections in the configuration file.
+
+1. `message_queue`: it contains the `address` and `port`, where the queue was deployed, or will be deployed with the
+   specified `credentials`. If the `ssh` property is presented, the Processing Broker will try to connect to
+   the `address` via `ssh` with provided `username` and `password` and deploy [RabbitMQ](https://www.rabbitmq.com/) at
```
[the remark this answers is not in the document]
tdoan2010: There is no default here. Users need to provide `username` and either `path_to_privkey` or `password`.

web_api.md
tdoan2010: I re-phrased it.

web_api.md
tdoan2010: I agree. The text was removed.

web_api.md
```diff
@@ -0,0 +1,261 @@
...
+The Web API specification can be found [here](openapi.yml). It follows
+the [OpenAPI specification](https://swagger.io/specification/). There are 4 parts to be implemented: discovery,
+processing, workflow, and workspace.
+
+**Discovery**: The service endpoints in this section provide information about the server. They include, but are not
+limited to, hardware configuration, installed processors, and information about each processor.
+
+**Processing**: Via the service endpoints in this section, one can get information about a
+specific [processor](https://ocr-d.de/en/spec/glossary#ocr-d-processor), trigger a processor run, and check a status of
+a running processor. By exposing these endpoints, the server can encapsulate the detailed setup of the system and offer
```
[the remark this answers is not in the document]
kba: OK, so you mean the processing broker can provide access to multiple processors? Can we formulate in such a way to make that clearer?

web_api.md
```diff
@@ -0,0 +1,261 @@
...
+    * `GET /processor/{executable}/{job-id}` for a single processor, or
+    * `GET /workflow/{workflow-id}/{job-id}` for the workflow.
+6. The result can be downloaded by calling the `GET /workspace/{workspace-id}` endpoint with the
+   header `Accept: application/vnd.ocrd+zip`. Without that header, only the metadata of the specified workspace is
+   returned.
+
+## Suggested OCR-D System Architecture
+
+There are various ways to build a system which implements this Web API. In this section, we describe a distributed
+architecture, which greatly improves the scalability, flexibility, and reliability of the system compared to
```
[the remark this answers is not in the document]
kba: Still open, right?

web_api.md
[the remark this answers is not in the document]
kba: OK, I would omit that.

```diff
-pre-processes it if necessary, and push it to an appropriate queue. A processing queue always has the same name as its
+pushes it to an appropriate queue. A processing queue always has the same name as 
```

web_api.md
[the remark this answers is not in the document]
kba: all good now, I would leave out the redundant "if necessary" at the end.

web_api.md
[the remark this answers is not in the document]
kba: But we do include the `/workspace` endpoint in the web api spec, don't we?

I think it would be beneficial for the IP reading this if you could add the information from [your comment](https://github.com/OCR-D/spec/pull/222#discussion_r1064891353) to the text here, so they get an idea of what is meant with "other storages" (better; "other storage solutions").

web_api.md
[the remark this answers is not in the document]
kba: I don't care much either way, it should just be consistent.

web_api.md
```diff
@@ -0,0 +1,186 @@
...
+      processor name and workspace ID, or
+    * Start a workflow on a workspace by calling the `POST /workflow/{workflow-id}` endpoint with the chosen workflow ID
+      and workspace ID.
+    * In both case, a job ID is returned to the user.
+5. With the given job ID, user can check the job status by calling:
+    * `GET /processor/{executable}/{job-id}` for a single processor, or
+    * `GET /workflow/{workflow-id}/{job-id}` for the workflow.
+6. The result can be downloaded by calling the `GET /workspace/{workspace-id}` endpoint with the
+   header `Accept: application/vnd.ocrd+zip`. Without that header, only the metadata of the specified workspace is
+   returned.
```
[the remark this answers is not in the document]
kba: Agreed.

web_api.md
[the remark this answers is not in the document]
tdoan2010: I gave some technologies used in this example in the paragraph direct under the figure.

bertsky (commented):
Is there any concept for a callback sub-API yet? (It's not a good design being forced to poll job status etc. For example, the spec could be extended with some `POST /callback` with parameters for the server to be registered and the kind of information requested)

bertsky:
Also: could we rename `Message Queue` → `Job Queue` or `Process Queue`? (Our system is not about messaging as such, msgqueue is an implementation detail.)

web_api.md
```diff
@@ -0,0 +1,265 @@
...
+and/or extend the reference implementation from OCR-D/core.
+
+**Workflow**: Beyond single processors, one can manage
+entire [workflows](https://ocr-d.de/en/spec/glossary#ocr-d-workflow), i.e. a series of connected processor
+instances. In this spec, a workflow amounts to a [Nextflow](https://www.nextflow.io/) script. Some information
+about Nextflow and how to use it in OCR-D is documented [in the Nextflow spec](nextflow).
+
+**Workspace**: The service endpoints in this section concern data management, which in OCR-D is handled
+via [workspaces](https://ocr-d.de/en/spec/glossary#workspace). Processing (via single processors or workflows) always
+refers to existing workspaces, i.e. combinations of `mets.xml` and downloaded files from the `mets.xml`.
```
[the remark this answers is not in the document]
tdoan2010: I didn't add this case since I want to leave this endpoint optional. Some IPs may want to support METS URL, some don't. And it also depends on where the METS files are stored, some kind of credentials might be necessary. That's why I leave it to the IP to decide.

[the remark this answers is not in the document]
bertsky: > I didn't add this case since I want to leave this endpoint optional. Some IPs may want to support METS URL, some don't. 

How about an **optional** endpoint then? IMO this should not be left underspecified (because it helps understanding the whole thing and because a standard endpoint would be preferable).

> And it also depends on where the METS files are stored, some kind of credentials might be necessary.

What do you mean by that?

tdoan2010:
> Is there any concept for a callback sub-API yet? (It's not a good design being forced to poll job status etc. For example, the spec could be extended with some `POST /callback` with parameters for the server to be registered and the kind of information requested)

To get notified about the job status, one can listen to the job result queue, as described in the `result_queue_name` part in the spec. So polling is not required.

We can also add the callback feature, but I think listening to the result queue is enough, or?

web_api.md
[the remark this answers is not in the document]
tdoan2010: Since I thought that all endpoints defined in the spec are compulsory. But maybe I can add some text to note that this endpoint is optional.

METS files are not always publicly accessible, right? There might be cases where authentication is needed to download the files.

bertsky:
> We can also add the callback feature, but I think listening to the result queue is enough, or?

Help me understand: so the queue is exposed as a Web API endpoint, too? Or is this the regular RabbitMQ API? (In that case, IMO it would help to give a brief explanation on how to use it in our spec, for that particular purpose.)

web_api.md
```diff
@@ -0,0 +1,265 @@
...
+   metadata of the workspace but not the files, simply set the header to `Accept: application/json`.
+
+## Suggested OCR-D System Architecture
+
+There are various ways to build a system which implements this Web API. In this section, we describe a distributed
+architecture, which greatly improves the scalability, flexibility, and reliability of the system compared to
+the [CLI](https://ocr-d.de/en/spec/cli) approach.
+
+<figure>
+  <img src="images/web-api-distributed-queue.jpg" alt="Distributed architecture with the Web API"/>
```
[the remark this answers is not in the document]
bertsky: It should show the client-server directionality though, not some dataflow logic.

web_api.md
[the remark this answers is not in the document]
bertsky: > Since I thought that all endpoints defined in the spec are compulsory.

The CLI spec has optional features, too.

> METS files are not always publicly accessible, right? There might be cases where authentication is needed to download the files.

Ah, ok. But why should that hinder establishing the endpoint itself? Even if an implementation needs to add credentials, or the URLs are not valid to some outside network, the server can still build on the same implementation.

tdoan2010:
No, it's the regular RabbitMQ API. The listener needs to connect to the RabbitMQ and listen to the result queue, as written in the spec:

> In case `result_queue_name` property is presented, the result of the processing will be pushed to the queue with the provided name. If the queue does not exist yet, it will be created on the fly. This is useful when there is another service waiting for the results of processing. That service can simply listen to that queue and will be immediately notified when the results are available.

For example, a job is submitted with `result_queue_name: ocrd-binarize-result`. Then, to get notified about the job status, your program needs to listen to the `ocrd-binarize-result` queue.

bertsky:
Thanks for the explanation!

> > In that case, IMO it would help to give a brief explanation on how to use it in our spec, for that particular purpose

Or perhaps provide an example of such a request to the RabbitMQ server in that paragraph?

tdoan2010:
I've added an example and some explanation in that paragraph. Please take a look :)

web_api.md
```diff
@@ -245,8 +245,47 @@ Instead of `path_to_mets`, one can also use `workspace_id` to specify a workspac
 from the Workspace Server. In case `result_queue_name` property is presented, the result of the processing will be
```
bertsky, line 255: Sorry, but I still don't understand. So the Processing Server (formerly Broker) receives some `POST /processor/{exe}` or `POST /workflow/{wid}` with parameters, respectively. It generates a job ID. It then divides the request up into a chain of processor jobs, creating a message to the queue for each one. Each queue is named by the processor executable.

- Is `result_queue_name` the same as the `appropriate queue` mentioned in the `Processing` paragraph of the architecture?
- How do the workers know when they have to skip some message because its prerequisite message has not finished, yet?

bertsky (commented):
Thanks!

bertsky:
> > Help me understand: so the queue is exposed as a Web API endpoint, too? Or is this the regular RabbitMQ API? (In that case, IMO it would help to give a brief explanation on how to use it in our spec, for that particular purpose.)

> No, it's the regular RabbitMQ API. The listener needs to connect to the RabbitMQ and listen to the result queue, as written in the spec:

It's still strange though that this callback mechanism should be so implementation/Python-specific, and also specific to the queue architecture. What if I just want a REST callback for when my single `POST /processor/{exe}` call finishes? What if the server in question does not use queues, but directly schedules local CLI calls or distributed Processor REST calls?

web_api.md
```diff
@@ -0,0 +1,318 @@
...
+There is normally no need to start (or stop) a Processing Worker manually, since it can be managed by a Processing
+Server via a [configuration file](#processing-server). However, if it is necessary to do so, there are two ways to start
+a Processing Worker:
+
+```shell
+# 1. Use ocrd CLI bundled with OCR-D/core
+$ ocrd processing-worker <processor-name> --queue=<queue-address> --database=<database-address>
+
+# 2. Use processor name
+<processor-name> processing-worker --queue=<queue-address> --database=<database-address>
```
bertsky, line 204: Are you sure about this mnemonic? It strikes me as inconsistent with the existing CLI conventions.

For the [REST wrapper](https://github.com/OCR-D/core/pull/884), you have added `--address`. Is this whole functionality obsolete now? (I thought that the reference implementation could easily provide both a worker for the queue model and a standalone REST wrapper along #884's lines, i.e. `POST /` etc. So maybe for the CLI something like `--server=worker --queue=... --database=...` and `--server=rest --address=...` could be defined.)

web_api.md
```diff
@@ -0,0 +1,318 @@
...
+```
+
+* `--queue`: a [Rabbit MQ connection string](https://www.rabbitmq.com/uri-spec.html) to a running instance.
+* `--database`: a [MongoDB connection string](https://www.mongodb.com/docs/manual/reference/connection-string/) to a
+  running instance.
+
+### Process Queue
+
+By using a queuing system, specifically [RabbitMQ](https://www.rabbitmq.com/), the reliability and flexibility of the
```
bertsky, line 213:
```suggestion
By using a queuing system for individual per-workspace per-job processor runs, 
specifically as message queueing with [RabbitMQ](https://www.rabbitmq.com/), the reliability and flexibility of the
```

web_api.md
```diff
@@ -0,0 +1,318 @@
...
+whole system are greatly improved. In our implementation, manual acknowledgement mode is used. This means, when a
```
bertsky, line 214: ...improved over **what**?

```suggestion
Processing Server are greatly improved over a system directly coupling the workflow engine and distributed processor instances.

 In our implementation of the Process Queue, manual acknowledgement mode is used. This means, when a
```

Would that be correct?

web_api.md
```diff
@@ -0,0 +1,318 @@
...
+3. `hosts`: this section contains a list of hosts, usually virtual machines, where Processing Workers should be
+   deployed. To be able to connect to a host, an `address` and `username` are required, then comes either `password`
+   or `path_to_privkey` (path to a private key). All Processing Workers, which will be deployed, must be declared under
+   the `deploy_processors` property. In case `type` is `docker`, make sure that [Docker](https://www.docker.com/) is
+   installed in the target machine and the provided `username` has enough rights to execute Docker commands.
+
+Among three sections, only the `message_queue` is required. However, if `hosts` is present, `mongo_db` must be there
+as well. For more information, please check the [configuration file schema](web_api/config.schema.yml).
+
+### Processing Worker
```
bertsky, line 193: I'd recommend moving this section below the Process Queue, because it is easier to explain top-down, and workers only make sense with queues.

web_api.md
```diff
@@ -0,0 +1,318 @@
...
+can be [overridden](https://www.rabbitmq.com/consumers.html#acknowledgement-timeout) by setting another value for
+the `consumer_timeout` property in the [`rabbitmq.conf`](https://www.rabbitmq.com/configure.html#config-file) file.
+
+To avoid processing the same input twice (in case of re-queuing), a Processing Worker first checks
+the [`redeliver`](https://www.rabbitmq.com/confirms.html#automatic-requeueing) property to see if this message was
+re-queued. If yes, and the status of this process in the database is not `SUCCESS`, it will process the data described
+in the message again.
+
+When a Processing Server receives a request, it creates a message based on the request content, then push it to an
+appropriate queue. A processing queue always has the same name as its respective processors. For
```
bertsky, line 228: What exactly does `appropriate queue` mean? Shouldn't this paragraph explain that a workflow must be decoded into single processor calls in a certain order, and that these calls then get pushed one by one onto the queue?

web_api.md
```diff
@@ -0,0 +1,318 @@
...
+parameters:
+  params_1: 1
+  params_2: 2
+result_queue_name: ocrd-cis-ocropy-binarize-result
+
+created_time: 1668782988590
+```
+
+In the message content, except `job_id`, `processor_name`, and `created_time`, are added by the Processing Server. The
+rest comes from the body of the `POST /processor/{executable}` request.
```
bertsky, line 252: I cannot decode this logically. Content minus job_id, or minus job_id and processor_name and created_time?

bertsky (commented):
more questions, some formulations

web_api.md
```diff
@@ -245,8 +245,47 @@ Instead of `path_to_mets`, one can also use `workspace_id` to specify a workspac
...
-and will be immediately notified when the results are available. The message schema for results can be
-found [here](web_api/result-message.schema.yml), while an example of a result message looks like this:
+and will be immediately notified when the results are available. Below is a simple Python script to demonstrate how a
+service can listen to the `result_queue_name` and act accordingly.
+
+```python
+import pika, sys, os
+
+def main():
+    connection = pika.BlockingConnection(pika.ConnectionParameters(host='localhost'))
```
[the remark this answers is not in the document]
tdoan2010: > In general, this would be the `--address` of the server.

No, it is the address of the RabbitMQ. The client connects directly to RabbitMQ to listen to the result queue, and it has nothing to do with Web API in this case.

> Also, what about credentials? The RabbitMq config seems to have username/password or keyfile. But again, how does the Web API client get to know?

Yeah, the client/listener must somehow know the credentials to be able to connect to RabbitMQ, maybe via the administrator, who configured RabbitMQ.

web_api.md
tdoan2010: > Is `result_queue_name` the same as the `appropriate queue` mentioned in the Processing paragraph of the architecture?

No, `appropriate queue` means the processing queues, which have the same name as the corresponding processor. For example, `ocrd-dummy` processors always listen to `ocrd-dummy` queue.

> How do the workers know when they have to skip some message because its prerequisite message has not finished, yet?

It doesn't know that. A worker is completely stateless. It only takes a message from a queue and start processing. It's the job of some kind of orchestrator above to push messages to queues in the right order.

tdoan2010:
> What if the server in question does not use queues, but directly schedules local CLI calls or distributed Processor REST calls?

That's a valid argument for having the callback feature. I will update the spec with this.

web_api.md
```diff
@@ -0,0 +1,318 @@
...
+which are Discovery Server, Processing Server, Workflow Server, and Workspace Server respectively. Although each server
+in the figure is deployed on its own machine, it is completely up to the implementors to decide which machines run which
+servers. However, having each processor run on its own machine reduces the risk of version and resource conflicts.
+Furthermore, the machine can be customized to best fit the processor's hardware requirements and throughput demand. For
+example, some processors need GPU computation, while others do not, or some need more CPU capacity while others need
+more memory.
+
+**Processing**: since the `Processing` section is provided by [OCR-D Core](https://github.com/OCR-D/core), implementors
+do not need to implement Processing Server, Process Queue, and Processing Worker themselves, they can reuse/customize
+the existing implementation. Once a request comes, it will be pushed to an appropriate queue. A processing queue always
```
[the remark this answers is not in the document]
tdoan2010: > But it says `compared to CLI` here. Shouldn't that rather be what benefits this particular architecture offers over a simple non-distributed Processing Server implementation (e.g. direct Processor REST calls)?

Agree. It's just the old text. I'll update it accordingly.

web_api.md
```diff
@@ -0,0 +1,318 @@
...
+has the same name as its respective processors. For example, `ocrd-olena-binarize` processors listen only to the queue
+named `ocrd-olena-binarize`. A Processing Worker, which is
+an [OCR-D Processor](https://ocr-d.de/en/spec/glossary#ocr-d-processor) running as a worker, listens to the queue, pulls
+new jobs when available, processes them, and push the job statuses back to the queue if necessary. One normally does not
+call a Processing Worker directly, but via a Processing Server. Job statuses can be pushed back to the queue, depending
```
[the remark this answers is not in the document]
tdoan2010: `call` means that users do not trigger Processing Workers directly (by sending requests to them).

web_api.md
```diff
@@ -0,0 +1,318 @@
...
+```shell
+$ ocrd processing-server --address=<IP>:<PORT> /path/to/config.yml
+```
+
+This command starts a Processing Server on the provided IP and port. It accepts only one argument, which is the path to
+a configuration file. The schema of a configuration file can be found [here](web_api/config.schema.yml). Below is a
+small example of how the file might look like.
+
+```yaml
+message_queue:
```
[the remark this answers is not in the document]
tdoan2010: Ah yes, I forgot to change it.

web_api.md
```diff
@@ -0,0 +1,318 @@
...
+  address: localhost
+  port: 5672
+  credentials:
+    username: admin
+    password: admin
+  ssh:
+    username: cloud
+    path_to_privkey: /path/to/file
+mongo_db:
```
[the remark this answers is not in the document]
tdoan2010: I just wanted to emphasize that it **must** be MongoDB. But using the key `database` is also fine for me.

web_api.md
```diff
@@ -0,0 +1,318 @@
...
+  address: localhost
+  port: 27017
+  credentials:
+    username: admin
+    password: admin
+  ssh:
+    username: cloud
+    password: "1234"
+hosts:
```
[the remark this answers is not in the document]
tdoan2010: Yes, it's a better term.

web_api.md
```diff
@@ -0,0 +1,318 @@
...
+2. `mongo_db`: this section also contains the `address` and `port`, where the [MongoDB](https://www.mongodb.com/) is
+   running, or will run. If `credentials` is presented, it will be used when deploying and connecting to the database.
+   The `ssh` section behaves exactly the same as described in the `message_queue` section above.
+3. `hosts`: this section contains a list of hosts, usually virtual machines, where Processing Workers should be
+   deployed. To be able to connect to a host, an `address` and `username` are required, then comes either `password`
+   or `path_to_privkey` (path to a private key). All Processing Workers, which will be deployed, must be declared under
+   the `deploy_processors` property. In case `type` is `docker`, make sure that [Docker](https://www.docker.com/) is
+   installed in the target machine and the provided `username` has enough rights to execute Docker commands.
+
+Among three sections, only the `message_queue` is required. However, if `hosts` is present, `mongo_db` must be there
```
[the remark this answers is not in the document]
tdoan2010: For the case when we have database and workers running 24/7 (OCR-as-a-Service) and just want to re-use them.

web_api.md
tdoan2010: > Is this whole functionality obsolete now?

For my initial idea, yes. I wanted to support only the queue architecture, but after reading your comments, I think it's better to have both, queue and standalone REST wrapper. The CLI syntax you suggested is a good idea.

web_api.md
tdoan2010: It is improved compared to other approaches, like distributed REST calls or local CLI calls.

web_api.md
tdoan2010: OK, fine for me.

web_api.md
tdoan2010: > Shouldn't this paragraph explain that a workflow must be decoded into single processor calls in a certain order, and that these calls then get pushed one by one onto the queue?

That's better. I'll re-write it in this way.

web_api.md
tdoan2010: Sorry, it should be a comma after "Processing Server", but somehow I put a dot. I will re-formulate it like this:

> In the message content, `job_id`, `processor_name`, and `created_time`, are added by the Processing Server, and the rest comes from the body of the `POST /processor/{executable}` request.

web_api.md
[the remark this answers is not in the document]
tdoan2010: > But it says `compared to CLI` here. Shouldn't that rather be what benefits this particular architecture offers over a simple non-distributed Processing Server implementation (e.g. direct Processor REST calls)?

Agree. It's just the old text. I'll update it accordingly.

web_api.md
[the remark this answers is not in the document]
bertsky: > > In general, this would be the `--address` of the server.
> 
> No, it is the address of the RabbitMQ. The client connects directly to RabbitMQ to listen to the result queue, and it has nothing to do with Web API in this case.

What I meant was: since the Processing Server usually deploys RabbitMQ locally, it would be the same host (and perhaps the default / hard-coded port). But otherwise, some mechanism of advertising the queue to all clients (hence `/discovery`) would be needed.

web_api.md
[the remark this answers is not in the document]
tdoan2010: I think `hosts` is a more appropriate term. On one host, there are many workers.

web_api.md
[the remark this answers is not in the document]
tdoan2010: Yes, but that is up to the IPs to decide, what they want to return from the `/discovery` endpoint. It's OK to return the address of the RabbitMQ instance, but credentials should not be returned, since the returned data is plain text.

So, in any case, the client must somehow know the address and credentials to connect to RabbitMQ, it can be via the administrator, as I said.

web_api.md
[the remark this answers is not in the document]
bertsky: Still – it's the name of the component in your architecture, of which the implementation is merely a detail. What if (at some later date) you want to replace MongoDB with something else? (You'd have to rename...)

web_api.md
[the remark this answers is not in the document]
bertsky: That's obvious though (multiple `address`  entries which contain multiple `processors` entries which may have `number_of_instance` >1 – these already indicate that there can be multiple workers per host entry).

However, IMHO what's not immediately clear, is what `hosts` stands for. That's too unspecific. And `worker_hosts` seems too long...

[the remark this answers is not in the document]
tdoan2010: Since I couldn't find a better term, I will just keep it `hosts` for the moment. I think the term is not too bad though 😅

web_api.md
tdoan2010: How about a syntax like this?

```shell
# Start a worker
$ ocrd server --type=worker --queue=... --database=...
$ ocrd-dummy server --type=worker --queue=... --database=...

# Standalone server
$ ocrd server --type=rest --address=... --database=...
$ ocrd-dummy server --type=rest --queue=... --database=...
```

web_api.md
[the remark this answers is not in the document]
bertsky: Ok, never mind.

web_api.md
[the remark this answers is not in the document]
tdoan2010: Since I didn't know how to detail describe the `credentials` part to cover all cases, I think it's best to leave it out. But never mind, as you said, we can have this endpoint defined and mark it as optional.

tdoan2010:
@bertsky could you resolve those conversations which are "done"? 😃 I know that it's still missing a section to describe the standalone REST processors, but I'll add it later, since I don't want to make this PR bigger (it's already big... 😅 ).

web_api.md
bertsky: Yes, I guess your cmd-group syntax is clearer. And it does help keep the standard help text short.

web_api.md
bertsky: Thanks – that's really helpful!

web_api.md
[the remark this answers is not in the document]
bertsky: Ok, granted, `/discovery` contents should not be overly specified. And since the queue model is optional, and you even have a callback endpoint now, and you amended the code snippet with port and credentials, it should be clear enough now.

web_api.md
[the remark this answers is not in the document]
bertsky: I'm still quite convinced these two arrows are in the wrong direction (renaming the arc to `listen to job status` does not change that – both kinds of request target the queue server and originate from the other side)

web_api.md
bertsky: Understood. Your additional terminology on job queue and result queue is also helpful, thanks.

But I still don't get the picture. So if there are multiple job queues (one for each available processor), but the workers are stateless, then how does the synchronisation between producers and consumers happen?

bertsky:
> @bertsky could you resolve those conversations which are "done"?  I know that it's still missing a section to describe the standalone REST processors, but I'll add it later, since I don't want to make this PR bigger (it's already big...  ).

Done. Three points remain open for me, hopefully we can still work these through.

web_api.md
tdoan2010: > So if there are multiple job queues (one for each available processor)

No, one queue for each type of processor. For example, 2 `ocrd-dummy` processors will listen to the same `ocrd-dummy` queue.

> how does the synchronisation between producers and consumers happen?

An example flow looks like this:

1. The producer, which in our case is the Processing Server, receives a `POST` request to the `/processor/ocrd-dummy` endpoint with this body:

```yml
path_to_mets: /path/to/mets.xml
input_file_grps:
  - OCR-D-DEFAULT
output_file_grps:
  - OCR-D-BIN
page_id: PHYS_001,PHYS_002
parameters:
  params_1: 1
  params_2: 2

result_queue_name: ocrd-dummy-result
callback_url: https://my.domain.com/callback
```

2. It creates a processing message based on this content and publish it to the `ocrd-dummy` queue.
3. An `ocrd-dummy` worker, which is listening to that queue, receives the message and starts processing.
4. When the processing finished, the worker saves the job status to the database, publishes a result message to the `result_queue_name` and also `POST`s the result message to the `callback_url`.

I hope that answers your question.

web_api.md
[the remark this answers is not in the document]
tdoan2010: OK, I updated the arrow direction 😃

tdoan2010:
> Done. Three points remain open for me, hopefully we can still work these through.

Sorry, but I only found 2 (the arrow direction and the sync between producers & consumers). What do I miss? 😅

web_api.md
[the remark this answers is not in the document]
bertsky: Thanks!

bertsky:
> What do I miss?

the one [about an optional `POST /workspace` for METS URLs](https://github.com/OCR-D/spec/pull/222#discussion_r1081556780) – since it's not in the swagger defs yet, and no mention of it being `OPTIONAL` in the markdown page, I assumed it's still open

web_api.md
bertsky: > > So if there are multiple job queues (one for each available processor)
> 
> No, one queue for each type of processor. For example, 2 `ocrd-dummy` processors will listen to the same `ocrd-dummy` queue.

That's what I meant. (Note I did not say _worker_ but _processor.)

> > how does the synchronisation between producers and consumers happen?
> 
> An example flow looks like this:

This example is too trivial – I already understood that much. What I meant by synchronisation between producers and consumers is that in every non-trivial workflow (more than 1 step), processors run one after another (for the same workspace), and each step (producing processor)'s output is the next step (consuming processor)'s input. IOW the consumer must not run before the producer.

Now, if the workers only get to see their own respective queue, and the steps are scattered across many queues – how is the workflow queued? (Or is there some runtime instance – like a thread for each workflow – that tracks progress, pushing new process messages to the job queues after receiving the previous result messages?)

tdoan2010: Ah, I see what you meant. In our current plan:

1. All processing workers are only consumers. It means that they won't publish any processing messages.
2. We need an orchestrator (Nextflow in our case) to execute a workflow. The orchestrator listens to the result queues and send the next `POST /processing/{exe}` request to the Processing Server to trigger the next processor when the previous one is finished.

For the next step, I'm thinking it might make sense to have a simple workflow definition file (YAML), in case Nextflow is overkill for any use cases. By reading that file, a processing worker knows what to do and which processor it has to trigger next when it finished. So in this case, a processing worker will also produce processing messages. But this is only the plan in my head. We need to finish what is defined in the current spec first.

tdoan2010:
> the one [about an optional `POST /workspace` for METS URLs](https://github.com/OCR-D/spec/pull/222#discussion_r1081556780) – since it's not in the swagger defs yet, and no mention of it being `OPTIONAL` in the markdown page, I assumed it's still open

I will update the `openapi.yml` file after this PR is merged. Then, that endpoint will be there and marked as `OPTIONAL`, as we discussed.

web_api.md
bertsky: Ah, of course, now I see (the wood from the trees). 

So perhaps the spec should also contain an example for the workflow server (detailling how the workflow gets decomposed into jobs by processor or even by page, and how the job/result queue communication then looks like)?

> For the next step, I'm thinking it might make sense to have a simple workflow definition file (YAML), in case Nextflow is overkill for any use cases. By reading that file, a processing worker knows what to do and which processor it has to trigger next when it finished. So in this case, a processing worker will also produce processing messages. But this is only the plan in my head. We need to finish what is defined in the current spec first.

Interesting idea.

web_api.md
```diff
@@ -0,0 +1,337 @@
+# Web API
+
+## Terminology
+
+* **Processing Worker**: a Processing Worker is an [OCR-D Processor](https://ocr-d.de/en/spec/glossary#ocr-d-processor)
+  running as a worker, i.e. listening to the Process Queue, pulling new jobs when available, processing them, and
+  pushing the updated job statuses back to the queue if necessary.
+* **Processing Server**: a Processing Server is a server which exposes REST endpoints in the `Processing` section of
+  the [Web API specification](openapi.yml).
+* **Process Queue**: a Process Queue is a queueing system for workflow jobs (i.e. single or chained processor runs on
```
bertsky, line 10:
```suggestion
* **Process Queue**: a Process Queue is a queueing system for workflow jobs (i.e. single processor runs on
```

web_api.md
```diff
@@ -0,0 +1,337 @@
...
+  one workspace) to be executed by Processing Workers. In our implementation,
+  it's [RabbitMQ](https://www.rabbitmq.com/).
```
bertsky, line 12:
```suggestion
  one workspace) to be executed by Processing Workers and to be enqueued by the Workflow Server via the Processing Server.
  In our implementation, it's [RabbitMQ](https://www.rabbitmq.com/).
```

web_api.md
```diff
@@ -0,0 +1,337 @@
+# Web API
+
+## Terminology
+
+* **Processing Worker**: a Processing Worker is an [OCR-D Processor](https://ocr-d.de/en/spec/glossary#ocr-d-processor)
+  running as a worker, i.e. listening to the Process Queue, pulling new jobs when available, processing them, and
+  pushing the updated job statuses back to the queue if necessary.
+* **Processing Server**: a Processing Server is a server which exposes REST endpoints in the `Processing` section of
+  the [Web API specification](openapi.yml).
```
bertsky, line 9:
```suggestion
* **Workflow Server**: a Workflow Server is a server which exposes REST endpoints in the `Workflow` section
of
   the [Web API specification](openapi.yml). In particular, for each `POST /workflow/{workflow-id}` request,
   a new task (in Nextflow: channel) comprised of a chain of `POST /processor/{executable}` jobs
   (in Nextflow: processes) is added and scheduled (communicating with the Processing Server for requests
   and the Process Queue for results).
* **Processing Server**: a Processing Server is a server which exposes REST endpoints in the `Processing` section of
  the [Web API specification](openapi.yml). In particular, for each `POST /processor/{executable}` request,
  a Processing Message is added to the respective Job Queue.
```

web_api.md
```diff
@@ -0,0 +1,337 @@
...
+in the figure is deployed on its own machine, it is completely up to the implementors to decide which machines run which
+servers. However, having each processor run on its own machine reduces the risk of version and resource conflicts.
+Furthermore, the machine can be customized to best fit the processor's hardware requirements and throughput demand. For
+example, some processors need GPU computation, while others do not, or some need more CPU capacity while others need
+more memory.
+
+**Processing**: since the `Processing` section is provided by [OCR-D Core](https://github.com/OCR-D/core), implementors
+do not need to implement Processing Server, Process Queue, and Processing Worker themselves, they can reuse/customize
+the existing implementation. Once a request arrives, it will be (split into single-processor jobs and) pushed to a
+job queue. A job queue always has the same name as its respective processors. For example, `ocrd-olena-binarize`
```
bertsky, line 114:
```suggestion
do not need to implement Workflow Server, Processing Server, Process Queue, and Processing Worker themselves, they can reuse/customize
the existing implementation. Once a workflow request arrives, it will be split into single-processor jobs
to be scheduled as processing requests whenever the previous job's result comes available. 
Each processing request will have a job pushed to the job queue named after the respective processor executable. For example, `ocrd-olena-binarize`
```

web_api.md
```diff
@@ -0,0 +1,337 @@
...
+**Database**: in this architecture, a database is required to store information such as users requests, jobs statuses,
+workspaces, etc. [MongoDB](https://www.mongodb.com/) is required here.
+
+**Network File System**: in order to avoid file transfer between different machines, it is highly recommended to have
+a [Network File System (NFS)](https://en.wikipedia.org/wiki/Network_File_System) set up. With NFS, all Processing
+Servers(specifically processors) can work in a shared storage environment and access files as if they are local files.
+To get data into the NFS, one could use the `POST /workspace` endpoint to
+upload [OCRD-ZIP](https://ocr-d.de/en/spec/ocrd_zip)files. However, this approach is only appropriate for testing or
+very limited data sizes. Usually, Workspace Server should be able to pull data from other storages.
+
```
bertsky, line 130: Perhaps add a 3rd level section on the Workflow Server (along the lines of the new entry in the terminology) here?

web_api.md
tdoan2010: As discussed, an example Nextflow script will be provided later.

web_api.md
tdoan2010: I would not mention the Workflow Server here, since we're talking about the `Processing` part of the API in this paragraph. And with the opening sentence:

> since the `Processing` section is provided by [OCR-D Core](https://github.com/OCR-D/core),...

adding the Workflow Server here doesn't fit.

web_api.md
tdoan2010: I will do it later, when we have the working example Nextflow script.

web_api.md
bertsky: Yes, I can see it more clearly now. But then the parenthesis (which from my suggestion) must be removed here, too. (Nothing needs to be split up, requests are already atomic. Except if splitting should happen here from document to page level...)

Perhaps the (split/monitor) relation between workflow requests and processing requests can be explained by a dedicated `Workflow` section then? (You do have the Workflow Server in the architecture figure and in the glossary...)

bertsky: Agreed!

bertsky:
Regarding https://github.com/OCR-D/spec/pull/222/files/ad6fe9eb018a40b823161331876f4afaad1827b3..3b2dee9d7c043f61ce04d3369b941f20ec78af2f – I do think sub-grouping is possible. It's been done in core in various places (e.g. ocrd-tool sub-group), too.

tdoan2010:
> Regarding https://github.com/OCR-D/spec/pull/222/files/ad6fe9eb018a40b823161331876f4afaad1827b3..3b2dee9d7c043f61ce04d3369b941f20ec78af2f – I do think sub-grouping is possible. It's been done in core in various places (e.g. ocrd-tool sub-group), too.

@kba said that it's really complicated to implement sub-command in the `OCR-D Click Wrapper`. So, better use `option`.

bertsky:
> @kba said that it's really complicated to implement sub-command in the `OCR-D Click Wrapper`. So, better use `option`.

I don't see it. It's just commands inside of groups inside of commands. A good example is ocrd-tool group. Implementation is easy. I could understand if the argument is that it's more complicated to document (esp. in the CLI). But then again, multiple related, interdependent options are also not easy not explain...

web_api.md
kba: Also, subcommands for the processor CLI are really tricky to do, so if it was to be based on the `ocrd-foo-processor` CLI, the worker arguments would realistically need to be options of the `ocrd-foo-processor` CLI, which makes the help text longer and requires more if-else-clauses to detect starting a worker vs. normal processing vs. `--help` etc.

So, I'm :+1: for a subcommand to `ocrd`.

bertsky: @kba perhaps you can comment on the respective [discussion in the Web API](https://github.com/OCR-D/spec/pull/222#issuecomment-1425934842) then.

kba:
> I don't see it. It's just commands inside of groups inside of commands. A good example is ocrd-tool group. Implementation is easy. I could understand if the argument is that it's more complicated to document (esp. in the CLI). But then again, multiple related, interdependent options are also not easy not explain...

The problem is that we generate the click interface programmatically, so that processor developers only need

```python
@click.command()                                                   
@ocrd_cli_options                                                  
def cli(*args, **kwargs):                                          
    return ocrd_cli_wrap_processor(MyProcessor, *args, **kwargs)
```

Now, we either require processors to change that to


```python
@click.group()
@ocrd_cli_options
def cli(...):
   return ocrd_cli_wrap_processor(...)

@cli.command('server')
def worker_cli(...):
  return ocrd_cli_worker_subcommand_wrap(MyProcessor)
```

or find a way to wrap this as a single annotation. It's possible but tricky.

Whereas, for the `ocrd` command, it's trivial to create subcommands because they use the standard click-way of creating CLI.

kba (approved):

PR #223, opened 2022-08-23
change content-type GET workflow/{workflow-id} according to POST /workflow

joschrew:
In `POST /workflow` (upload a workflow-script) the content-type is `text/vnd.ocrd.workflow`. Because of that I think the content type to get back a workflow must be the same and not `application/vnd.ocrd+zip`.

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -219,7 +219,7 @@ paths:
           description: Return Workflow
           content:
             application/json: {schema: {$ref: '#/components/schemas/Workflow'}}
-            application/vnd.ocrd+zip: {}
+            'text/vnd.ocrd.workflow': {}
         '404':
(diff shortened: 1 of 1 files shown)


kba (approved):

joschrew:
Reminder (rather for myself) to know what I mean:
`POST /workflow` is used to upload a workflow. There the content-type for the request body is  `text/vnd.ocrd.workflow`. I assume this is right.
`GET /workflow/{workflow-id}` us used to retrieve a workflow. There the content-type of the response is `application/vnd.ocrd+zip`. But it returns a workflow and not a workspace so I think the content-type has to be changed to one mentioned before. I think this is just a copy-and-paste leftover, where the creator forgot to change the type of the response.

PR #232, opened 2022-11-23
cli: multi-value values cannot contain comma, OCR-D/core#921

kba:


cli.md
```diff
@@ -11,7 +11,8 @@ because [`-p` is repeatable](#-p---parameter-param_json).
 
 **NOTE**: Parameters marked **MULTI-VALUE** cannot be repeated but can specify
 multiple values, formatted as a single string with comma-separated items (e.g.
-`-I group1,group2,group3` instead of `-I group1 -I group2 -I group3`).
+`-I group1,group2,group3` instead of `-I group1 -I group2 -I group3`). Due to
+comma being used as a separator, the values cannot contain commas themselves.
```
bertsky, line 15:
```suggestion
comma being used as a separator (and not allowed in METS identifiers), 
the values themselves must not contain commas.
```

bertsky (approved):
Perhaps also:

```
**NOTE**: For parameters which allow for a combination of multi-value, range and/or regex,
the operator precedence is  **MULTI-VALUE** > **REGEX** > **RANGE**.
```

kba:
> Perhaps also:
> 
> ```
> **NOTE**: For parameters which allow for a combination of multi-value, range and/or regex,
> the operator precedence is  **MULTI-VALUE** > **REGEX** > **RANGE**.
> ```

b84c20e

PR #233, opened 2022-12-06
ocrd_tool: clarify the description of "size" for resources

kba:


kba:
I think this can be merged, it's a small change to make it clear that `size` always refers to what is retrieved over the network.

PR #236, opened 2022-12-19
QA Spec - Schema

kba:
Just the (still evolving) schema, so #225 can focus on the spec itself.


ocrd_eval.sample.json
```diff
@@ -0,0 +1,120 @@
...
+      },
+      "workflow_model": "Fraktur_GT4HistOCR",
+      "document_metadata": {
+        "fonts": [
+          "antiqua",
+          "fraktur"
+        ],
+        "publication_century": "1800-1900",
+        "publication_decade": "1850-1860",
+        "publication_year": 1855,
```
mweidling, line 38: Currently, Matthias' [GT metadata tool](https://tboenig.github.io/gt-metadata/document-your-gt.html) allows for providing an upper and lower bound for the publication date which I translate to `publication_century`. Do we need a more granular date in QuiVer (or any other tool we might use this JSON schema for)? If not, I suggest to settle with `publication_century` for now. 

What do you think, @kba?

mweidling:
`processing_time`  as  benchmark for a page is only given in the example files and not defined in the schema YAML. Since it's a bit tricky to calculate it I also wanted to know what the overall value of this is. If it doesn't have that much of a value I recommend to toss it and focus on `pages_per_minute` instead.

@kba What do you think?

ocrd_eval.schema.json
```diff
@@ -0,0 +1,238 @@
...
+                }
+              }
+            ]
+          }
+        }
+      }
+    },
+    "EvaluationMetrics": {
+      "cer": {
+        "description": "CER calculated over the text of a whole page (in by_page) or combined text of all pages (in document_wide)"
```
mweidling, line 204: If we look at `document_wide`: Is there a difference between `cer` and `cer_mean`?

mweidling: Same applies for `by_page`.

bertsky: IMO you have to drop the `cer` entry. It's `cer_mean` / `cer_median` / `cer_stddev` / `cer_range`, regardless of the hierarchy level.

ocrd_eval.schema.yml
```diff
@@ -0,0 +1,192 @@
...
+        description: Human readable name and version of evaluation tool used (for UI
+
+      document_metadata:
+        type: object
+        title: Bibliographical and typographical metadata about the work to be evaluated
+        properties:
+
+          publication_year:
+            type: number
+            description: Year he document was originally published
```
mweidling, line 103:
```suggestion
            description: Year the document was originally published
```

ocrd_eval.schema.yml
```diff
@@ -0,0 +1,192 @@
...
+
+          publication_century:
+            type: string
+            description: Century he document was originally published
```
mweidling, line 107:
```suggestion
            description: Century the document was originally published
```

ocrd_eval.schema.yml
```diff
@@ -0,0 +1,192 @@
...
+            type: array
+            items:
+              type: string
+              enum: ['antiqua', 'fraktur']
+
+      provenance:
+        type: object
+        description: Information on which tools in which version were used in determining metrics
+        properties:
+          paramters:
```
mweidling, line 133:
```suggestion
          parameters:
```

ocrd_eval.schema.yml
```diff
@@ -0,0 +1,192 @@
...
+        properties:
+
+          publication_year:
+            type: number
+            description: Year he document was originally published
+
+          publication_century:
+            type: string
+            description: Century he document was originally published
+            pattern: '[12][0-9]{3}-[12][0-9]{3}'
```
mweidling, line 108:
```suggestion
            pattern: '[12][0-9]{3}-[12][0-9]{3}'
```
What about documents before 1000? Or do we restrict the schema to prints?

mweidling: Also, the code should be more like

```suggestion
            pattern: '[12][0-9]00-[12][0-9]00'
```
Since we focus on centuries. It's rather uncommon to have a century beginning at e.g. 1856.

ocrd_eval.schema.yml
```diff
@@ -0,0 +1,192 @@
...
+
+          publication_decade:
+            type: string
+            description: Decade the document was originally published
+            pattern: '[12][0-9]{3}-[12][0-9]{3}'
```
mweidling, line 113:
```suggestion
            pattern: '[12][0-9]{2}0-[12][0-9]{2}0'
```
See comment(s) above.

ocrd_eval.schema.yml
```diff
@@ -0,0 +1,192 @@
...
+            type: string
+            description: Description of this workflow step
+
+      workflow_model:
+        type: string
+        description: Human readable name of the main model used for recognition in the OCR workflow (for UI)
+
+      eval_tool:
+        type: string
+        description: Human readable name and version of evaluation tool used (for UI
```
mweidling, line 94:
```suggestion
        description: Human readable name and version of evaluation tool used (for UI)
```

ocrd_eval.sample.json
kba: I think `publication_year` would be enough, decade and century are derived from it.

ocrd_eval.schema.json
kba: `cer` without qualifiers would be document-wide on the whole text, whereas the `_mean`, `_median` etc. suffixes indicate aggregate values over the pages.

But that's just my initial idea, if you think we don't need a global CER value for a whole-document, we can drop it.

ocrd_eval.schema.yml
kba: I was thinking of incunabula where we do have a fairly specific start date (1452-1454). But if you prefer "clean" centuries from `xx00-yy00`, that's also fine with me.

kba:
> `processing_time` as benchmark for a page is only given in the example files and not defined in the schema YAML. Since it's a bit tricky to calculate it I also wanted to know what the overall value of this is. If it doesn't have that much of a value I recommend to toss it and focus on `pages_per_minute` instead.
> 
> @kba What do you think?

That was an oversight that it's not in the schema. Why is it tricky to calculate or rather, how would you calculate `pages_per_minute` without it?

ocrd_eval.schema.json
bertsky: I don't understand. These two aspects (granularity and statistic) should be completely independent.

`EvaluationReport` has `document_wide` and `by_page`, each of which can use all of `EvaluationMetrics`.

What is `cer` other than `cer_mean`?

ocrd_eval.schema.yml
mweidling: I don't have a strong opinion about this topic. If what you propose is what is most common in the target domain, then we go for that.

ocrd_eval.sample.json
mweidling: Alright :)

ocrd_eval.schema.json
mweidling: I share @bertsky 's thoughts (that's what my initial question aims at).

mweidling: Ping @kba

ocrd_eval.schema.yml
```diff
@@ -0,0 +1,192 @@
...
+        allOf:  [{ '$ref': '#/$defs/LabeledUrl' }]
+        description: The workspace containing the GT
+
+      workflow_steps:
+        type: object
+        description: Human readable description of the individual steps in the workflow (for UI)
+        patternProperties:
+          '^[0-9]+$':
+            type: string
+            description: Description of this workflow step
```
mweidling, line 86:
```suggestion
        type: array
        description: Human readable description of the individual steps and their parameters in the workflow (for UI)
        minItems: 1
        items:
          type: object
          properties:
            id:
              type: string
              description: The name of the processor used for this workflow step
              pattern: '^ocrd-[a-z\-]+'
            params:
              type: object
              description: A map of parameters and their values applied to the processor used for this workflow step
          required: ['id', 'params']
```

ocrd_eval.schema.yml
```diff
@@ -0,0 +1,192 @@
...
+
+          layout:
+            type: string
+            enum: ['simple', 'complex']
+
+          fonts:
+            type: array
+            items:
+              type: string
+              enum: ['antiqua', 'fraktur']
```
mweidling, line 127:
```suggestion
              enum: ['Antiqua', 'Black Letter']
```

mweidling:
> > `processing_time` as benchmark for a page is only given in the example files and not defined in the schema YAML. Since it's a bit tricky to calculate it I also wanted to know what the overall value of this is. If it doesn't have that much of a value I recommend to toss it and focus on `pages_per_minute` instead.
> > @kba What do you think?
> 
> That was an oversight that it's not in the schema. Why is it tricky to calculate or rather, how would you calculate `pages_per_minute` without it?

My first naive approach would be to look at the processing time for the complete work, calculate the mean processing time for a page and use this as basis for obtaining  `pages_per_minute`. The arithmetic mean for a page cannot be used for `processing_time`, though, as this would imply the actual time this page took.

ocrd_eval.schema.yml
mweidling: These are the values I conveniently get from the data, but we could stick to the draft, too. I just need to know.
Is `fraktur` used more widely than `Black Letter`, btw?

ocrd_eval.schema.json
kba: Agreed, let's drop `cer` without qualifier.

ocrd_eval.schema.yml
kba: `Black letter` is equivalent to "gebrochene Schriften", of which "Fraktur" is but one - used in prints rather than manuscripts like others such as "Textura" or "Bastarda", cf. https://github.com/OCR-D/ocrd_typegroups_classifier#colwise-classifier

So "Black Letter" is more general but in this case, we do actually mean "Fraktur" AFAIK the GT data (@tboenig - do we have non-Fraktur black letter in our data?).

ocrd_eval.schema.json
kba: 30f1436

kba:
> > > `processing_time` as benchmark for a page is only given in the example files and not defined in the schema YAML. Since it's a bit tricky to calculate it I also wanted to know what the overall value of this is. If it doesn't have that much of a value I recommend to toss it and focus on `pages_per_minute` instead.
> > > @kba What do you think?
> > 
> > 
> > That was an oversight that it's not in the schema. Why is it tricky to calculate or rather, how would you calculate `pages_per_minute` without it?
> 
> My first naive approach would be to look at the processing time for the complete work, calculate the mean processing time for a page and use this as basis for obtaining `pages_per_minute`. The arithmetic mean for a page cannot be used for `processing_time`, though, as this would imply the actual time this page took.

`pages_per_minute` is likely the more interesting measure, so feel free to focus on that.

ocrd_eval.schema.yml
mweidling: Thanks for elaborating!

In the METADATA.yml we have

```yaml
description: >-
  Ground Truth with a collection of documents with the following
  characteristics: font blackletter
```
and 

```yaml
script:
  - Goth
```
I re-checked the way the font classification is done and realized that I could just as easily use `Fraktur`. So let's go for that if @tboenig doesn't have any objections.

mweidling:
> > > > `processing_time` as benchmark for a page is only given in the example files and not defined in the schema YAML. Since it's a bit tricky to calculate it I also wanted to know what the overall value of this is. If it doesn't have that much of a value I recommend to toss it and focus on `pages_per_minute` instead.
> > > > @kba What do you think?
> > > 
> > > 
> > > That was an oversight that it's not in the schema. Why is it tricky to calculate or rather, how would you calculate `pages_per_minute` without it?
> > 
> > 
> > My first naive approach would be to look at the processing time for the complete work, calculate the mean processing time for a page and use this as basis for obtaining `pages_per_minute`. The arithmetic mean for a page cannot be used for `processing_time`, though, as this would imply the actual time this page took.
> 
> `pages_per_minute` is likely the more interesting measure, so feel free to focus on that.

Let's skip `processing_time` for pages then for now. I'll make an issue so that we can implement it in a later stage.

ocrd_eval.schema.yml
```diff
@@ -0,0 +1,217 @@
...
+      by_page:
+        type: array
+        description: Metrics page-by-page
+        items:
+          type: object
+          allOf: [
+            { $ref: '#$defs/CommonEvaluationMetrics' },
+            { $ref: '#$defs/PageId' }
+          ] 
+          unevaluatedProperties: false
```
mweidling, line 167: This might not be the best solution, but it's the only one I managed so far to achieve a strict validation of the properties.
Open to suggestions!

kba: > > `unevaluatedProperties` is like `additionalProperties`, except that it can "see through" `$ref` and "see inside" `allOf`, `anyOf`, `oneOf`, `if`, `then`, `else`, and whatever else I'm forgetting right now.

(https://github.com/json-schema-org/json-schema-spec/issues/556)

I use `additionalProperties: false` all the time and IIUC `unevaluatedProperties` offers the same but can traverse referenced subschemas, so I think this is actually is the best solution.

mweidling: thanks for the feedback!

ocrd_eval.schema.yml
kba: @tboenig is for fine-grained font description with the list of values from https://tboenig.github.io/gt-guidelines/html/trans/lySchriftarten.html

mweidling: No problem, I can add that.

mweidling: See [00a9020](https://github.com/OCR-D/spec/pull/236/commits/00a9020b02dafeb32b77f7512644bfe21e1f881b). Resolved?

kba:
Sry, I accidentally rebased instead of merging, but it is all in master now.

mweidling:
> Sry, I accidentally rebased instead of merging, but it is all in master now.

Alright, thank you!

mweidling:
I'll remove the branch then.

PR #244, opened 2023-03-24
fix img src and max-height

lena-hinrichsen:


kba (commented):
Why, what is the probelm with `<figure/>`?

lena-hinrichsen:
> Why, what is the probelm with `<figure/>`?

The class does not neccessarily to be changed, as it seems. that happened when I copy pasted a block from another page (`about`) where we use `figure img`. 

The important change is the image src, max-height and the href I think

kba (approved):
OK, I was just curious, looks legit otherwise, feel free to publish

kba (changes requested):
Did not realize this is OCR-D/spec not OCR-D/website. I need to adapt `make spec` in OCR-D/website, please don't merge for now.

PR #245, opened 2023-03-24
Dockerfile: improve and update to GHCR

bertsky:


docker.md
```diff
@@ -1,28 +1,41 @@
...
+to facilitate implementation of new and integration of existing tools)
+to handle recurrent tasks in a spec-conformant way.
+
+Moreover, this makes using natively installed and containerized [CLI](cli) interchangeable.
 
 ## Naming images
 
-Image tags MUST be the same as the project name but with underscore (`_`)
+Docker image tags MUST be the same as the project name but with underscore (`_`)
 replaced with forward slash (`/`).
```
kba, line 32:
```suggestion
replaced with forward slash (`/`) and `ocr-d` spelled in lower case.
```

kba (approved):
LGTM, thanks!

docker.md
bertsky: Is that the right context to prescribe `ocr-d` though? IIUC this spec applies to any external module, too. Within github.com/OCR-D we will use that prefix sure, but others should be free to do their thing (including Dockerhub or GHCR or some Gitlab or whatever registry), right?

kba: Sure, I only wanted to prevent people from running into the same problem with `ocrd vs OCR-D vs ocr-d`. In DockerHub I could add people to the team "ocrd", so they could publish their repos (which did not need to be in the `OCR-D` organization in GH) to `docker.io/ocrd/calamari` for example.

Since that does not work with ghcr.io AFAIK, we should maybe not restrict the naming of the image at all?

bertsky: > because that's forbidden, see `docker-image-tag(1)`

Actually, it's not forbidden to use underscores in image tags, only in the host name part of the registry. (I did not catch this when I first read it, because the documentation is a bit cryptic.)

And because for external projects (outside of github.com/OCR-D), their connection to OCR-D is _only_ expressed with the typical `ocrd_` prefix, I would even say we should remove this formulation entirely, or at least soften it from a `MUST` to a `SHOULD`. 

For example, it does not make sense to remove the `ocrd_` from [bertsky/ocrd_detectron2](https://github.com/bertsky/ocrd_detectron2/): `bertsky/detectron2` would be confused with a variant of [detectron2](https://github.com/facebookresearch/detectron2) proper, and `bertsky/ocrd/detectron2` is not allowed.

(I am not saying we should completely reverse course and even use `ocrd/ocrd_tesserocr`, although I would not object to it. But mandating the removal of `ocrd` and underscores just does not make sense for external modules.)

kba (approved):

PR #262, opened 2025-03-11
ocrd-tool schema: no default for parameter 'content-type'

bertsky:
(as the default makes no sense for non-file parameters, i.e. everything other than `type=string format=uri`)


bertsky:
(arguably, the same goes for `cacheable`...)

MehmedGIT (approved):

kba (approved):

PR #263, opened 2025-03-12
ocrd_tool: processor resource resolution, and update...

bertsky:
- rename `file parameters` to `resource parameters` (matching `cli#processor-resources` )
- add multi-valued semantics to `content-type` via comma separated string
- stay neutral regarding file vs. directory, replace `fpath` with `rpath`
- no more recursive relative names, except for CWD (!)
- introduce (i.e. proper define/list) resource locations
- update fileGrp cardinality instead of names
- update schema verbatim
- update and change to simpler example (ocrd_tesserocr→ocrd_kraken)
- improve formulations

### DIFF
--- a/ocrd_tool.md
+++ b/ocrd_tool.md
@@ -10,50 +10,61 @@ services](swagger).
 
 To validate a `ocrd-tool.json` file, use `ocrd ocrd-tool /path/to/ocrd-tool.json validate`.
 
-## File parameters
+## Resource parameters
 
-To mark a parameter as expecting the address of a file, it MUST declare the
-`content-type` property as a [valid media
-type](https://www.iana.org/assignments/media-types/media-types.xhtml).
-Optionally, workflow processors can be notified that this file is potentially
-large and static (e.g. a fixed dataset or a precomputed model) and should be cached
-indefinitely after download by setting the `cacheable` property to `true`.
+To mark a parameter as expecting the address of a file or directory,
+it MUST declare the `content-type` property as a [valid media
+type](https://www.iana.org/assignments/media-types/media-types.xhtml),
+which may contain _multiple values_, formatted as a single string with comma-separated items.
 
-The filename itself, i.e. the concrete value `<fpath>` of a file parameter,
-should be resolved in the following way (with `<cwd>` representing the current 
-working directory, and `<mod>` representing the distribution directory of the module):
+Optionally, workflow processors can be notified that this file or directory
+is potentially large and static (e.g. a fixed dataset or a precomputed model),
+and should be cached indefinitely after download, by setting the `cacheable` property
+to `true`.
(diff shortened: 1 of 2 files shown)


bertsky:
Moreover, we have to consider whether we really want to keep the first 2 lines:
https://github.com/OCR-D/spec/blob/fcefac1ba789bdd123edd5d210c6b8662aa246d4/ocrd_tool.md?plain=1#L28-L29

According to https://github.com/OCR-D/core/issues/782 we never actually implemented this.

bertsky:
Regarding the last commit, do we want to include the newly defined environment variables in the CLI spec? And what about the `worker` and `server` subcommands?

MehmedGIT:
> Regarding the last commit, do we also want to include the newly defined environment variables in the CLI spec? 

For completeness, yes, why not?

> And what about the `worker` and `server` subcommands?

Honestly, not sure. The assumption is that the Processing Server config file will be used to deploy network agents. The `worker` and `server` subcommand options trigger the executables over the SSH, and they were not meant to be used separately. The same is true with the `resmgr-server` subcommand. 

Since the `ocrd network client ...` subcommand will be actively used by users, it makes sense to document only that command.
However, we are addressing the `spec` and not just user documentation, it still makes sense to describe all network-related subcommand. Except the processor-related `worker` and `server` subcommands. I.e., `ocrd network processing-worker [OPTIONS] executable` is fine, but `executable [worker|server] [OPTIONS]` feels outside the core scope.

ocrd_tool.md
```diff
@@ -10,50 +10,61 @@ services](swagger).
...
-    * Split the variable value at `:` and try to resolve by appending `<fpath>`
-      to each token and return the first found file if any
-  * `$XDG_DATA_HOME/ocrd-resources/<name-of-processor>/<fpath>` (with `$HOME/.local/share` instead of `$XDG_DATA_HOME` if unset)
-  * `/usr/local/share/ocrd-resources/<name-of-processor>/<fpath>`
-  * `<mod>/<fpath>` (**Note** that the file is expected to be directly under `<mod>`, not in a subdirectory)
+The path of the file or directory itself, i.e. the concrete value `<rpath>` of a 
+resource parameter, should be resolved in the following way:
+
+* If `<rpath>` is an `http:` or `https:` URI, download to a temporary directory
+  (if `cacheable==False`) or a semi-temporary cache directory (otherwise).
```
kba, line 29: Have we implemented this?

kba (approved):
Many thanks, very thorough, clarifications overdue.

ocrd_tool.md
bertsky: see [above](https://github.com/OCR-D/spec/pull/263#issuecomment-2718528648)

PR #264, opened 2025-04-17
ocrd_tool: allow version to contain extra tag like `.post1` or `.rc2`

kba:
We append a suffix for our local deployments.

With this PR we use the convoluted but complete regex recommended to parse [python packaging compliant version strings](https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers-regex)


bertsky (approved):
LGTM

PR #265, opened 2025-09-02
Remove processor server

joschrew:
When planning Ocrd-Network, we considered two types of processors to use: the Processing Worker and the Processor Server. In [pull request 468](https://github.com/OCR-D/ocrd_all/pull/468) in ocrd all the Processor Server is not used but still ocrd-network offers the whole capability like the previous native installation. In addition, the Processing Worker provides largely the same functions as the Processor Server, making the Processor Server largely redundant. 
Removing the Processor Server simplifies the code and reduces the effort required for maintenance and understanding the code.


kba (approved):
Thanks! As discussed last week, I'm fine with removing the process**or** server.

